PR #32, opened 2016-11-25
Bugfixing and enhancements

v1r0x:
This started as a small bugfix branch, but had to add a couple of other fixes and enhancements.
Here's a short list of things which didn't work (because they were buggy or not implemented);

1. Add markers using the leaflet panel
2. Add/Remove source entries
3. Store dimensions/epochs
4. Removed unnecessary dummy elements and other small UI enhancements
5. Open datepicker popup
6. Store data not only in the database but local as well (no refresh of the site after adding/editing/removing data from an element)

Please review and give your feedback **ASAP** or merge.
@dirkseidensticker @derntl @AngusLeBoeuf 
Thanks in advance!

### DIFF
--- a/controllers/mainCtrl.js
+++ b/controllers/mainCtrl.js
@@ -97,11 +97,19 @@ spacialistApp.controller('mainCtrl', ['$rootScope', '$scope', 'scopeService', 'h
         });
     };
 
+    var getStoredQueries = function() {
+        httpGetFactory('../spacialist_api/analysis/queries/getAll', function(queries) {
+            console.log(queries);
+            $rootScope.storedQueries = queries;
+        });
+    };
+
     var updateInformations = function() {
         getContexts();
         getArtifacts();
         getMarkerChoices();
         getLiterature();
+        getStoredQueries();
     };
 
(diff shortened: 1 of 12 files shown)


dirkseidensticker:
Testing revealed one issue within the sources/literature-window:
* Adding literature doesn't work (no error message in the console)
* the button for adding additional references is not active either)

v1r0x:
>     Adding literature doesn't work (no error message in the console)
>     the button for adding additional references is not active either

Adding literature works fine for me. Could you enable the Network tab in e.g. firefox' dev console? Maybe there is an error in a GET or POST request.

The button is only active if you've selected a literature and entered a text.

dirkseidensticker:
Thanks for verification. Works as it should. I wasn't aware that it's necessary to enter a text as well. Sorry for the hold up.

v1r0x:
We should add an information text that it is necessary to add a text or allow adding literature without text.

dirkseidensticker:
Yeah, that would be helping. I'll add it to our ToDo's. Cheers

PR #60, opened 2017-01-30
User management

v1r0x:
Finally!
The already existing token authentication plugin is now configured and active. Beside that I implemented a user management backend.
The authentication is done using tokens no matter which role or permission a user has. A role is just a permission collection and all checks ("Is the user allowed to ...") are based on permissions. This allows us to have _one_ user interface which hides/displays elements based on permissions and every project can define it's own roles. I added a couple of roles for testing purposes. Everyone of @eScienceCenter/spacialists is admin and can add/remove/edit (currently only password) users and all the other stuff. If you want to test another role please add a new user and add the desired role(s).

Please test and review @eScienceCenter/spacialists


v1r0x:
Please note that this is a rather preconfigured implementation. In the future we should extend the user management to allow admins to create/delete new roles. Then we could remove the predefined roles and let the users decide which role they need (beside the admin role).

PR #80, opened 2017-02-07
Fix photo tab bugs

### LINKED ISSUE #74: Photos-Tab
- [x] Local testing installations that point to the same database are unable to obtain any images.
- [x] Upload of images doesn't work; after selecting an image from local disk nothing happens.
- [x] Setting a link between an image and a context doesn't work (through right click and then click on 'Mit aktuellem Kontext verbinden' isn't working

v1r0x:
Fix #74 

I tested it on our testing environment and the bugs should be fixed now. Could you please test it on your VMs @dirkseidensticker and @AngusLeBoeuf ?


dirkseidensticker:
Doesn't work in the VM. getAll-Method fails

v1r0x:
Works fine on our testing machine. Maybe there is problem with your php or permissions.

dirkseidensticker:
It couldn't work on local testing environment because: lumen searches for those images within the local files path. How should be deal with this? (together with @v1r0x )

PR #84, opened 2017-02-08
Fix storing of elements

### LINKED ISSUE #75: Saving of Attributes doesn't work
The first error message (`setMarkerOption is not defined`) get's thrown when setting an attribute (e.g. selecting something from a dropdown list). When clicking 'Speicher' nothing happens either and another error (`httpPostPromise is not defined`) appears.

v1r0x:
Fix #75 


lumen/app/Http/Controllers/ContextController.php
```diff
@@ -665,37 +563,26 @@ public function updateOrInsert($request, $cid, $isUpdate, $user) {
...
-                                    ['context_id', '=', $attr->context_id],
-                                    ['attribute_id', '=', $attr->attribute_id],
-                                    ['id', '=', $attr->id]
-                                ])
-                            ->update($data);
+                        $attrValue = AttributeValue::where([
+                            ['context_id', '=', $attr->context_id],
+                            ['attribute_id', '=', $attr->attribute_id],
+                            ['id', '=', $attr->id]
+                        ]);
```
dirkseidensticker: should be `])->first();`

PR #93, opened 2017-02-15
Unset sources as well

v1r0x:
In the bibliography tab the source entries from the previous selected element are still visible if the user deselects an element.


v1r0x:
This is another problem. If you select an element in the tree (an element with literature references), switch to the 'Quellennachweise' tab and then deselect the element, there are still the references from the previous selected element.

Issue #95, opened 2017-02-15
after login the UI needs a reload


derntl:
because it is empty otherwise

v1r0x:
This may be fixed with one of the open PRs. At least I don't have a blank page.
Does this always happen or only in a specific case?

derntl:
seems so 🤞

PR #103, opened 2017-02-15
display error messages with same status code only once

### LINKED ISSUE #94: have to click 5 error boxes away before logging in

v1r0x:
Fix #94 
Tried to fix issue #95 as well, but couldn't figure out how to tell the ui-router plugin to reload the page (controllers and services). I tried `$state.go('spacialist', {}, { reload: true });` and `$state.reload('spacialist');`. Maybe we have to fix it different...

### DIFF
--- a/app.js
+++ b/app.js
@@ -618,6 +618,20 @@ spacialistApp.config(function($controllerProvider, $provide) {
 
 spacialistApp.config(function($stateProvider, $urlRouterProvider, $authProvider, $httpProvider, $provide) {
     var lastError;
+    var rejectReasons = [
+        'user_not_found',
+        'token_not_provided',
+        'token_expired',
+        'token_absent',
+        'token_invalid'
+    ];
+    var rejectTranslations = [
+        'Invalid username or password',
+        'Authentication error (token not provided)',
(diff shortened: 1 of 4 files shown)


v1r0x:

v1r0x:
We could display an error message above or below the login form instead of a popup. Of course only if the error is one of (user_not_found, token_expired, token_invalid, token_absent, token_not_provided). Thus, the user could still see the reason why they have to login again, but we could get rid of the popup.

derntl:
Yes that would be okay, although I don't think it is needed. In any case, I think;
* instead of alert-danger, I would use alert-info class.  
* No popup window
* Use non-tech language, avoid the use of the word "token". A typical end user will not care about auth tokens. Instead of "token not provided" a message could be "You are not logged in. Please log in." , or instead of "token expired" the message should be "Your login has expired. Please log in again."

v1r0x:
I replaced the popup window with an infobox below the login form. I (tried to) replace(d) the lumen error messages with less techy ones. They can be found [here](https://github.com/eScienceCenter/Spacialist/blob/login-fix/app.js#L628-L634). Feel free to edit them.

Please review again @derntl

PR #136, opened 2017-03-02
Popup behaviour

v1r0x:
This should hopefully fix the wrong behaviour of the marker popups. @derntl could you please test it again? I'm not sure if I tested all possibilites.

I also altered the marker layout:
- The button to link a geodata object with the current object is now bigger.
- The buttons (link and unlink) contain the context's name (e.g. 'Unlink from Context XY')
- The heading (name of the context) is only visible if the context is linked to the geodata object. I think it's confusing if the context name is the heading of the object's popup although it's not connected to it.


v1r0x:
I merged the latest changes from master branch into this one and can not reproduce any of your issues (your list in this issue and the list in #141 ).
Could you please test this again @derntl ?

derntl:
I approved this now, but we need to rethink the user experience :-)

PR #147, opened 2017-03-03
Image upload

### LINKED ISSUE #142: no photo upload confirmation
when i upload a photo there is no confirmation when and whether it worked. only expanding the "Gesamte Fotosammlung" panel shows whether it has worked.

v1r0x:
Fix #142 

Due to a service/controller problem the progress bars were not visible. Now there should be a progress bar for each file. After all files are uploaded the progress bars are removed (after 1000ms). Is this ok or should there be more/another info? @derntl


v1r0x:
The problem is that our lumen part uploads the images to the local storage (the machine php is running on), but to the database specified in the `.env`. Thus images uploaded on our server are stored on the server (and are displayed if you open the .../intern link) and images uploaded on your local instance are stored on your machine (and are displayed if you open localhost/...).

PR #149, opened 2017-03-06
remove legacy update marker code

### LINKED ISSUE #148: context name change fails
changing the name of a context throws `Error: $scope is not defined` and achieves nothing

v1r0x:
Fix #148 

The problem was a reference to old code. Changing the context name should now be possible without an error.


v1r0x:
Please review again @derntl

PR #159, opened 2017-03-10
add warning message if there are unsaved element changes

### LINKED ISSUE #154: inconsistent need for saving changes
**Situation**
* Some actions change the DB without any need to click "Save" or anything (e.g. linking a context with a marker or deleting a child context)
* Edits in the context form are discarded without notice when the user does not click "Save" explicitly
* Edit of the context name is reflected in the tree and map popup as the user types, but if "Save" is not clicked, the change is not persisted in the DB. In fact, until the next reload, the GUI suggests that the changes were stored in the DB by showing the unsaved edited name in tree, form and map.

**Some possible remedies**
* notify the user that leaving the form discards the edits and ask whether to save or discard (my preferred way)
* consistently save every change without explicit Saving (can be problematic without undo functionality, a box I think we should not open)

v1r0x:
Fix #154 

This PR enhances the tree. If the user adds/deletes/edits the properties of an element and clicks on another element, deselects an element or clicks on a geodata object a warning popup is displayed. The user has three options:
1. Discard changes and switch to the other element or whatever the user wanted to do
2. Save changes and then complete the users action
3. Cancel

I'm still looking for a solution for the problem that a user can click on a popup and the popup opens in the background independent of the users choice (1., 2. or 3.). But you can test the current state and give some feedback @eScienceCenter/spacialists


dirkseidensticker:
Adding new Objects like contexts and artifacts through right-click is still possible. This should also trigger the warning popup.

v1r0x:
I added the check for unsaved changes to the right-click/context menu and the 'create new top-context' button as well. Could you please test again, @dirkseidensticker ?

PR #161, opened 2017-03-13
close previous popup if unlinked element is selected & was linked

### LINKED ISSUE #151: 1:1 mapping between marker and context
see discussion in #150

v1r0x:
Fix #151 

It should now no longer be possible to link a geodata object to multiple contexts (popup should close if it is connected to a context).


dirkseidensticker:
Got it; work's as advertised 🎱

PR #164, opened 2017-03-15
set max-zoom to appropriate value (27 = ~1.5ft)

### LINKED ISSUE #163: Zoom factor in the map not sufficient
Especially with the geodata of the Lukanien project and with respect to future projects that include excavations it might be important to be able to zoom further in. Currently a zoom factor with a scale of 30 m is the maximum. Once we have even more precise data (Tachymetry of finds within cm accuracy) this won't be sufficient.

v1r0x:
Fix #163 
max zoom level is now at 27, which is about 1.5-2ft (less than 1m) on the map.

### DIFF
--- a/app/mapService.js
+++ b/app/mapService.js
@@ -219,6 +219,9 @@ spacialistApp.service('mapService', ['httpGetFactory', 'httpPostFactory', 'httpG
     function initMapVariables() {
         map.map = {};
         map.map.center = {};
+        map.map.defaults = {
+            maxZoom: 27
+        };
         map.map.bounds = map.createBoundsFromArray([
             [-90, 180],
             [90, -180]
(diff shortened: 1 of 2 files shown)


derntl:
fixed max zoom is not so optimal, some base maps go gray after certain zoom levels. so i guess this setting should be configurable

v1r0x:
I think it is better (at least easier) to have a fixed value. If a layer doesn't support a certain zoom level, the users don't need it (at this zoom level), anyway. With the fixed value they have the chance to use a wider range of zoom levels, but don't have to. They can still stop at a certain level and don't press the '+' button or scroll.

v1r0x:
Good point. Could discuss this again if we come up with a good solution for project configuration.

PR #201, opened 2017-04-04
Photo fixes

### LINKED ISSUE #188: photo fixes
- [x] delete phtos
- [x] change properties of photos
- [x] filter by catch prases or folder structure to organize pics, plans, drawing...
- [x] notification that more pics are available then the seven displayed

v1r0x:
Fix #188 

This PR adds more functionality to the photos tab.
- It is now possible to delete photos (from DB and storage)
- Show a placeholder if more images are available
- Search (See comment below)
- Edit properties (not yet implemented)

Please review/comment @eScienceCenter/spacialists


v1r0x:
I added a basic search bar, but I'm not sure what we really need. The current solution works only with tags as `OR`.

Example:
1. Image with tag _sun_
2. Image with tag _holiday_

If the user adds _sun_ as search term, image 1. is displayed, but not image 2. If the user also adds _holiday_ to the search term, both images are displayed.

Is this the intended behaviour or do we only want to display images that have both tags (`AND`)?

dirkseidensticker:
Concerning the search bar there should be three independent search field:

* Tags (as there are now)
* Date (created)
* Camera model

v1r0x:
I added the missing parts (edit photo properties and filter). The filters for camera model and the date (created) are `OR` filters, tag filter is `AND`. `AND` doesn't make sense for camera and date, since one photo can only have one camera and one date :wink:. Should we switch from a multiple choice to a single choice for camera and date? I think it could be useful if the user searches for e.g. images with tags _Befundfoto_ **AND** _Objektfoto_ which were taken on _20.3.2016_ **OR** _21.3.2016_ with camera _Nikon D7000_ **OR** _Canon Rebel T3i_. What do you think? @eScienceCenter/spacialists and especially @dirkseidensticker

PR #210, opened 2017-04-06
get rid of remaining font awesome icons

derntl:
plus, the "create top context" is now a button

### DIFF
--- a/includes/tree.html
+++ b/includes/tree.html
@@ -1,14 +1,14 @@
 <div ui-tree data-nodrop-enabled="true" data-drag-enabled="false" id="context-tree">
     <ul ui-tree-nodes ng-model="contexts" style="margin-bottom: 8px;">
-        <li class="clickable" ng-click="createNewContext()">
-            <i class="material-icons placeholder"></i><i class="material-icons md-18">add</i>
+        <li class="btn btn-primary clickable" ng-click="createNewContext()">
+            <i class="material-icons md-18">add</i>
             <span class="contextPrefix">{{'context-tree.new-toplevel-context'|translate}}</span>
         </li>
         <li ng-repeat="id in contexts.roots" ng-show="contexts.data[id].visible" ui-tree-node data-collapsed="true" ng-include="'includes/tree-child.html'" context-menu="setContextMenu">
         </li>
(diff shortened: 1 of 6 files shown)


includes/tree.html
```diff
@@ -1,14 +1,14 @@
 <div ui-tree data-nodrop-enabled="true" data-drag-enabled="false" id="context-tree">
     <ul ui-tree-nodes ng-model="contexts" style="margin-bottom: 8px;">
-        <li class="clickable" ng-click="createNewContext()">
-            <i class="material-icons placeholder"></i><i class="material-icons md-18">add</i>
+        <li class="btn btn-primary clickable" ng-click="createNewContext()">
```
v1r0x: The buttons are way too big and have too much padding. What's the problem with simple `<li>` elements?

layouts/add-attribute.html
```diff
@@ -9,11 +9,11 @@
...
                     <div class="input-group">
                         <input type="text" class="form-control" ng-model="mc.label" ng-model-options="{ debounce: 250 }" placeholder="Search" uib-typeahead="result as result.label for result in mc.onSearch($viewValue)" typeahead-min-length="3" typeahead-wait-ms="50" typeahead-loading="loadingResultsLabel" typeahead-no-results="noResultsLabel">
-                        <span class="input-group-addon"><i class="fa fa-fw fa-search"></i></span>
+                        <span class="input-group-addon"><i class="material-icons">search</i></span>
                     </div>
-                    <i ng-show="loadingResultsLabel" class="fa fa-fw fa-refresh"></i>
+                    <i ng-show="loadingResultsLabel" class="material-icons">refresh</i>
                     <div ng-show="noResultsLabel && mc.label.length >= 3">
-                        <i class="fa fa-fw fa-times"></i> No Results Found
+                        <i class="material-icons md-18">block</i> No Results Found
```
v1r0x: I prefer the clear icon. Additionally that's the icon they use in their examples. Any preference?

includes/tree.html
derntl: as discussed, this element triggers a data changing action and should therefore be button styled

layouts/add-attribute.html
derntl: alright ⛏

PR #214, opened 2017-04-10
Drag & drop

### LINKED ISSUE #46: Changing hierarchy positions of elements in the tree
There should be the option to change then hierarchical position of a find or context later on. Best solution seems to be a Drag and Drop function comparable to the one in ThesauRex.

v1r0x:
Fix #46 

This PR adds drag & drop support to Spacialist. I also had to update the delete, edit and duplicate, so please check them as well. Duplicate by the way now inserts the duplicate after the duplicated element instead of the end.


v1r0x:
This is related to the tree plugin. In my understanding it works this way:
1. Remove dragged element from source (in your example the root elements array)
2. On drop calculate the index of the drop position

Since they remove the element in the first place, they calculate the drop index based on the wrong array.
I could implement some kind of _workaround_, but I don't know if it is worth it. The users can still move it up again. What do you think?

PR #218, opened 2017-04-11
Sets margins and padding to 0

dirkseidensticker:
At the moment a lot of space is wasted within the forms. This change sets all the marings and the padding to 0 to get as much information displayed as possible.

What do you think?

Please review and comment (but **don't merge**) @eScienceCenter/spacialists 

virtualbox_ubuntu 16 04_11_04_2017_10_57_57_modds

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -635,8 +635,8 @@ ul.nav .material-icons, button .material-icons, ul.dropdown-menu .material-icons
     cursor: pointer;
     font-size: 12px !important;
     position: absolute;
-    right: 1px;
-    top: 2px;
+    right: -2px;
+    top: 8px;
 }
 
 .list-group-item {
--- a/css/style.css
+++ b/css/style.css
@@ -95,9 +95,8 @@ textarea {
     resize: vertical;
 }
 
-.form-group {
-    margin: 0 0 0 0 !important;
-    margin-bottom: 0 !important;
+.form-group-unpadded {
+    margin: 0 !important;
     padding-bottom: 0 !important;
 }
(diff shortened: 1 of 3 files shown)


v1r0x:
I'd add at least 1 or 2px of margin/padding. Maybe we should add our own class `form-group-unpadded`, so we can decide if we want to overwrite the padding/margin of a specific `form-group. element

AngusLeBoeuf:
I think the "new" version with margin/padding 0 looks better than the old. At some points it seems to be a little too compressed, maybe we can test some other values between the two extremes.

AngusLeBoeuf:
Ok, I checked some other values and in my opinion the changes made by @dirkseidensticker are fine and we should keep them. With this change there are some minor problems in the photo tabs. @v1r0x could you please check this?

v1r0x:
The filter form in the photo section needs some polishing too.

derntl:
Passiert hier noch was oder kübeln wir den PR?

dirkseidensticker:
I've created a new css class names `form-group-unpadded` that sets all margins and padding to 0. This class was then used for the forms in questions.

Additionally I tried to place the label attributes as well as the star inside a `table-cell`, in order prevent overlap between these two elements. Unfortunately this resulted in some irregular behavior regarding the vertical alignment.

@eScienceCenter/spacialists Please review and suggest improvements

v1r0x:
I modified the label layout in `inputFields.html` and the results look good. Could anyone else please have a look at it? @eScienceCenter/spacialists

PR #229, opened 2017-05-08
File viewer

v1r0x:
This PR extends the current photo viewer to a file viewer.
The user can now upload every type of file (we should discuss if we want to allow it or what restrictions we want).
It also includes a viewer for plain text files (e.g. code, html, csv, ...) and highlighting (for code, html, ...) and a pdf viewer. The pdf viewer is very simple and still a bit buggy (zooming can lead to a wider pdf than the container).

**This is still WIP** but please test and review it and tell me what you think about file restrictions @eScienceCenter/spacialists 

Please run `bower update` and `php artisan migrate`


derntl:
I think one should be able to upload anything, but of course there'll be no preview for some files.

dirkseidensticker:
The filters should be collapsible , like the other entries. But unlike the rest, the filters should stay open if opened once.

v1r0x:
I added a 3D library which currently supports Collada (`.dae`) and soon OBJ (`.obj`) files (including `.mtl`). In the 3D viewer the user can zoom and measure the length between two points on the object (click on the ruler to enable/disable the measurement mode and then select two points on the object's surface). Unfortunately the calculated length is only displayed after the user clicks on the ruler again or on one of the zoom buttons.

v1r0x:
There you go, @dirkseidensticker

PR #235, opened 2017-05-23
fix context type label typo

derntl:
How beautiful this looks now 😍 

image


v1r0x:
Is the `<br>` worth it? IMO we only decrease the form height without getting better UX/UI. If we want to keep the `<br>` I'd suggest something like that (or the other way round or remove the properties heading)
form-header

derntl:
looks lika a tag cloud. but i don't care as long as the context type is visible somewhere near the name of the element

PR #237, opened 2017-05-30
Dynamic UI

v1r0x:
This PR allows the user to resize the different columns on the different pages (currently only main page).
The buttons to extend/shrink a column are displayed (for all columns) if the user hovers over one of the columns.
The size of the columns is currently limited to values from 1 to 11 (grid size is 12, 0 is unfortunately not supported).
Please review @eScienceCenter/spacialists (please note that this is still WIP)


derntl:
Great that one can adjust the columns now.

One problem: if you scroll in the form or in the tree, the resize control in moves along with the scrolling.

Personally I do not really like the visual appearance (huge buttons, always visible), but that is personal taste. I would put the buttons at the very bottom, max 30 x 30 each, and only visible when the mouse is hovering over them.

v1r0x:
> Personally I do not really like the visual appearance (huge buttons, always visible), but that is personal taste. I would put the buttons at the very bottom, max 30 x 30 each, and only visible when the mouse is hovering over them.

My idea was to hide/show them if the user is in an edit mode or "normal" mode. And yes, the css is still a bit buggy. It should only display the buttons of the hovered container, not of all.

derntl:
[Fenster ein/ausblenden und in der Größe verschieben (dynamisches UI)](https://trello.com/c/sh8l3dzZ/131-fenster-ein-ausblenden-und-in-der-grosse-verschieben-dynamisches-ui)

v1r0x:
The buttons are now only visible if the user has enabled the edit mode **and** is hovering over a resizeable div (only the particular div has the buttons).
The scrolling thing is still not working properly. I fixed the scrolling itself, but now it adds 100-200px above the parent div (100 if it only has one extend button, 200 if it has both buttons). Don't know how to set a fixed position and ignore scrolling.

dirkseidensticker:
I like the possibility of resizing the width of the columns as well. Some observations:
- The third column gets thrown beneath the first once the second column is fully expanded towards the left. This is a bit counter intuitive.
- The same behavior can be observed once you shrink the first column too much. Then everything else get's tossed downwards.

v1r0x:
> The third column gets thrown beneath the first once the second column is fully expanded towards the left. This is a bit counter intuitive.
> A similar behavior can be observed once you shrink the first column too much. Then everything else get's tossed downwards.

This is because bootstrap relies on a 12 column grid. If a row is wider than the maximum size of 12 columns, it breaks the content at the container which exceeds the 12th column. In our case this happens if one of the three column blocks is shrinked to 0. Then they get a fixed width of 100px (to display the buttons) and thus it is 12 columns + 100px.
One possible solution would be to move the buttons to the adjacent container.
Example: To extend the left container the user has to click on a button on the left side of the middle container and to shrink the right container the user has to click on a button on the left side of the right container.
If a container gets shrinked to zero they don't need a temporary 100px container, because they can be extended again using a button in the adjacent container.

> The Headings and everything beneath them get are moved downwards once you activate the Edit-Mode.

I couldn't find a solution where the buttons are aligned in the middle of a column and this was the best I could get. Feel free to waste another couple of hours trying to fix it :wink:

PR #251, opened 2017-07-07
Fix search

v1r0x:
This PR fixes two small issues reported from one of the projects.
1. [x] Childrens of search results in the tree view are not visible (only the expand arrow)
2. [x] _Show tags_ in the filter tab in the files tab has no effect (thus it is not possible to add tags to files)


v1r0x:

Issue #238, opened 2017-05-31
Global search


derntl:
Enable global free text search, e.g. via a search bar in the header.

This is probably functionally dependent on #189

v1r0x:
Since #189 is fixed and #272 is merged, we can move this one back to Delphi.

PR #272, opened 2017-08-17
Switch to UI-Router

### LINKED ISSUE #189: Allow specify some target in URL
Sharing a URL to some specific thing in Spacialist (e.g. the bibliography section, role management, or some selected context or marker) should be allowed.

v1r0x:
Fix #189, base for #238 
This PR is not ready to merge since there are a couple of bugs/limitations. See the list below for an overview. If you encounter any bug or weird behaviour during testing, please add a comment to this PR. Please test @eScienceCenter/spacialists 

## Known Bugs/Problems
- [x] _Save & Continue_ in the unstaged changes popup doesn't work
- [x] Sources & certainty in the certainty popup are stored, but only displayed after a page refresh
- [x] Linked files are not displayed
- [ ] After drawing a new geo object, the selected context is no longer selected (by design? bug?)
- [x] Editing marker properties doesn't work (e.g. color or lat/lng)
- [x] Linking contexts to geodata objects is not possible and may be falsely restricted
- [x] Unlinked geodata has empty popup and displays empty context detail page
- [x] When one opens a context detail page (`/context/<id>`) directly, the popup of the linked geo object doesn't get opened. When one opens a link to a geo object (`/geodata/<gid>`), the page stays empty due to several errors.
- [x] Validate states in `onEnter` function (e.g. `return $state.target('root.spacialist');` if validation fails)

## Beautify
- [x] Replace unused fields/methods
- [x] Replace unnecessary views in db (already done in `user-prefs` branch)
- [x] Rename controllers/services/etc. (e.g. image/photo => file)

## External deps
- [x] Edit/delete geodata (depends on Leaflet.draw)
- [x] Selecting a context in the tree sometimes displays the wrong context in the detail page (after necessary switch to npm #276, seems to be a bug in the sticky-state plugin) **disabled sticky for now**


v1r0x:
Functionality should be almost done. Could you please all test this branch and have a look at the open bugs/questions in OP? Thanks! @eScienceCenter/spacialists

v1r0x:
I fixed most of the remaining bugs/todos. There are still things that don't work 100%, but nothing critical. Maybe something for a 0.3.1 release :wink: 
Could please anyone test and confirm this so we can finally release Cannae? Thanks! @eScienceCenter/spacialists

PR #278, opened 2017-09-18
Add (user) prefs

### LINKED ISSUE #119: Store user & project preferences in the DB
Personal prefs should be persisted in the DB, including:
- [x] when a user changes the GUI langauge
- [x] when a user changes the column layouts

Project wide prefs should also be configurable and persisted in the DB:
- [x] the concept URL in thesaurus which is used as the root concept for tagging files in Spacialist
- [x] whether or not the geo map (or other part of the UI) shall be displayed. 
- [x] whether or not a link/button to the ThesauRex instance should be displayed
- [x] whether or not tooltips should be shall be displayed
- [x] Project name (displayed in the upper left corner and in ThesauRex URIs)

v1r0x:
Fix #119 

I implemented the prefs stated in #119. For the "Go to ThesauRex" button I used a simple checkbox. If checked, a link to `./thesaurex` is added to the header. I already discussed the option to set a string (which could also be a absolute path to another server) instead of a checkbox with @dirkseidensticker. An empty string would be equal to an unchecked box, every non-empty string would be used as link target. What's your opinion @eScienceCenter/spacialists ?


v1r0x:
Fixed both of your requests. Please test again @derntl

PR #285, opened 2017-09-27
Polish file modal

v1r0x:
This PR fixes missing meta data, the TODO in the _Linkages_ tab and introduces a basic exif data viewer.
Due to the hack/workaround for the file modal it is not possible to switch the state to a context (from the linkages tab). This will be fixed in a later commit in this PR.


v1r0x:
> Due to the hack/workaround for the file modal it is not possible to switch the state to a context (from the linkages tab). This will be fixed in a later commit in this PR.

This is fixed now. Please test @eScienceCenter/spacialists

PR #317, opened 2017-12-18
Version info

v1r0x:
This PR adds an about section to the main menu. The release information is taken from a local file (`lumen/app/.VERSION`). I added a `pre-commit` hook to my cloned repo, but there are two problems with that approach:
1. Hooks are only local. Thus, everybody has to copy the hook to their cloned repo and check for updated versions.
2. I simply run `git describe --tags > lumen/app/.VERSION` and `git log -1 --format=%at >> lumen/app/.VERSION` and then append the changes to the current commit. Because this is a pre-commit hook, it returns the version info **before** the current commit
3. Updates that are made directly on GitHub do not update the version, because hooks are local

I added a fallback which runs the commands in PHP if the file does not exist. The problem with that solution is, that it only works for cloned installs. If someone downloads a release as zip, they don't get any version information.

Easiest solution would be to manually create the version file, when a release is made. This would fix the missing version info for downloaded zip installs and uses the current fallback for all cloned installs.
What's your opinions @eScienceCenter/spacialists ?

about


derntl:
Great! 

I would say we worry about that when someone who downloads the software and complains why there is no release info. To some degree of certainty this will never happen ;-)

PR #321, opened 2017-12-31
changed the logout button further right

dirkseidensticker:
The tools and settings menus seem to me crammed between the logout button and the language settings. I would suggest to place the logout botton further right, maybe even to the right of the helper icon. Like that the UI features the two important drop-down menus more prominently.

### DIFF
--- a/templates/header.html
+++ b/templates/header.html
@@ -35,11 +35,6 @@
                 </li>
             </ul>
             <ul class="nav navbar-nav navbar-right">
-                <li ng-if="$ctrl.user.user.name">
-                    <a href="" ng-click="$ctrl.logoutUser()">
-                        <i class="material-icons">power_settings_new</i> {{ $ctrl.user.user.name }}
-                    </a>
-                </li>
                 <li class="dropdown" uib-dropdown>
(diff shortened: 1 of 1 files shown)


v1r0x:
IMO the order should be (from left to right):
1. Logout
2. Settings/Tools
3. Tools/Settings
4. Language
5. Help Icon

dirkseidensticker:
@v1r0x which is exactly the current layout. While using it yesterday I found it more important and intuitive if the tools and settings are the first things within the menu. People are interacting more with those I think.

v1r0x:
I was confused, because you branched from master. The menu in the master branch is messed up. Ephesus has the "correct" (from my point of view) layout.

I'd keep it that way, but if others prefer your solution, I'm ok with it.
**BUT**: This branch is branched from master, not the ephesus release. You should either close this one and re-branch from the latest ephesus tag or move it to the Federsee milestone.

v1r0x:
Please comment @derntl @AngusLeBoeuf

derntl:
In most apps I know the logout/profile button is the right-most button. But I do not think we should spend time on cosmetics right now. We can keep it in mind and merge something like this later.

derntl:
I'm closing this for now.

PR #329, opened 2018-01-30
Update en.json

derntl:
transalte two German strings


v1r0x:
This should go into 0.5.2, but your branch is branched from master. Please open a new one which is branched from 0.5.2. Thanks!

PR #330, opened 2018-01-31
Extend context-type type handling

v1r0x:
Based on project feedback, I removed the differentiation between _context_ and _find_. The project now has to decide which of the context-types can be a top-level context-type and which context-types are allowed as sub-contexts of a particular context-type in the hierarchy.

You have to run migrations.

### TODO
- [x] Drag & Drop
- [x] After adding a new context-type, the form is empty (reload required)
- [x] Add (De)Select-All buttons in DME


v1r0x:
Should be fixed now

PR #331, opened 2018-02-21
changed color to background for elemnts in the tree

dirkseidensticker:
As elements in the tree are sometimes unreadable due to bright colors and we use background colors in ThesauRex I tried to change the tree in Spacialist to use background colors as well.

screenshot 2018-02-21 12 19 04

Possible ToDo's:
- [ ] switch text color to white if the`main.getColorForId` return a darker value. I've changed `var lgt` to compensate, but that value might be changed back to `'65%'`.

@eScienceCenter/spacialists please review & discuss

### DIFF
--- a/includes/tree-child.html
+++ b/includes/tree-child.html
@@ -10,7 +10,7 @@
     <span ui-sref="root.spacialist.context.data({id: r.id })" ui-sref-active="selected-leaf-child" style="display: table-cell; padding-left: 4px;">
         <i class="material-icons md-18" aria-hidden="true" ng-if="contexts.data[r.id].type == 0">account_balance</i>
         <i class="material-icons md-18" aria-hidden="true" ng-if="contexts.data[r.id].type == 1">spa</i>
-        <span ng-style="options.getColorForId(contexts.data[r.id].uri)" ng-bind-html="contexts.data[r.id].name | sphighlight:highlight"></span> <span ng-if="contexts.data[r.id].uri" class="contextPrefix">{{ concepts[contexts.data[r.id].uri].label}}</span>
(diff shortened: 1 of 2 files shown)


derntl:
I think it's a good idea!

v1r0x:
Good idea :+1:

includes/tree-child.html
```diff
@@ -10,7 +10,7 @@
     <span ui-sref="root.spacialist.context.data({id: r.id })" ui-sref-active="selected-leaf-child" style="display: table-cell; padding-left: 4px;">
         <i class="material-icons md-18" aria-hidden="true" ng-if="contexts.data[r.id].type == 0">account_balance</i>
         <i class="material-icons md-18" aria-hidden="true" ng-if="contexts.data[r.id].type == 1">spa</i>
-        <span ng-style="options.getColorForId(contexts.data[r.id].uri)" ng-bind-html="contexts.data[r.id].name | sphighlight:highlight"></span> <span ng-if="contexts.data[r.id].uri" class="contextPrefix">{{ concepts[contexts.data[r.id].uri].label}}</span>
+        <span ng-style="options.getColorForId(contexts.data[r.id].uri)" style="padding: 0 5px; border-radius: 3px" ng-bind-html="contexts.data[r.id].name | sphighlight:highlight"></span> <span ng-if="contexts.data[r.id].uri" class="contextPrefix">{{ concepts[contexts.data[r.id].uri].label}}</span>
```
v1r0x: I'd move the styling to it's own class. And add some pixel of top padding to "center" the text.

v1r0x:
Since the new version does not have colors in the tree at all, I'd close this PR. If you want (these) colors back, please open a new issue, so we can add this to the next release.

@eScienceCenter/spacialists @dirkseidensticker do you agree?

dirkseidensticker:
I'm okay with closing that PR. I'll create an new Issue if background colors should be brought back.

PR #357, opened 2020-03-05
Group restrictions

v1r0x:
This adds a concept similar to roles to the core. But instead of restricting access to a whole part of Spacialist (entities, files, ...) access can be restricted for each item individually (and their sub-items). A restriction can be one of the following three rules:
1. Write access (same as without this feature)
2. Read-only access (same as not having the "delete and move concepts" and "duplicate and edit concepts" permissions)
3. No access
Currenty only entities are supported. 

### View for user with read-only access
group-restrictions-readonly

### User management
group-restrictions-users

### View of access menu
group-restriction-added
group-restrictions-add


v1r0x:
This is outdated. Starting a new approach in https://github.com/DH-Center-Tuebingen/Spacialist/tree/0.10-working-groups

v1r0x:
See #440

PR #424, opened 2023-10-27
[0.10] Add serif font

v1r0x:


v1r0x:
@Severino should we merge `0.10-style` as well?

PR #425, opened 2023-10-27
[0.10] Add datatype Richtext

v1r0x:


v1r0x:
Used wrong base branch...

PR #428, opened 2023-11-07
[0.10] [WIP] Upgrade test suite to latest versions

v1r0x:


v1r0x:
@Severino Workflow itself good so far. Setup is fine. PHP tests fail, because they are outdated. We can work on this as soon as `0.10` is (almost) ready to release

PR #429, opened 2023-11-07
0.10 refactor vite

Severino:
Replaced laravel mix with vite as a build tool.


INSTALL.md
```diff
@@ -185,10 +185,10 @@ If you want to send mails to your users, you have to adjust the `MAIL_*` setting
 ### Configure JavaScript
 Spacialist is based on several JavaScript libraries, which are bundled using Webpack (configuration is done using Laravel Mix, a webpack-wrapper for Laravel). Only the zipped releases contain the already bundled JavaScript libraries. All other users have to run webpack to bundle these libraries.
```
Severino, line 186: This section is outdated now.

package.json
```diff
@@ -38,10 +37,12 @@
         "eslint": "^8.51.0",
         "eslint-plugin-vue": "^9.17.0",
         "laravel-mix": "^6.0.49",
```
Severino, line 39: laravel-mix should be removed

package.json
```diff
@@ -14,11 +14,9 @@
     },
     "homepage": "https://github.com/DH-Center-Tuebingen/Spacialist#readme",
     "scripts": {
-        "dev": "npm run development",
-        "development": "npx mix",
-        "watch": "npx mix watch",
-        "prod": "npm run production",
-        "production": "npx mix -p",
+        "dev": "vite",
```
v1r0x, line 17: We should also add a dev build for `IS_OPEN=true`

vite.config.js
```diff
@@ -0,0 +1,26 @@
...
+// import react from '@vitejs/plugin-react';
+import vue from '@vitejs/plugin-vue';
+import { fileURLToPath, URL } from 'url';
+
+const isOpen = process.env.IS_OPEN === 'true';
+
+export default defineConfig({
+    plugins: [
+        laravel([
+            (isOpen) ? 'resources/js/app.js': 'resources/js/open.js',
```
v1r0x, line 12: This has to be either `(!isOpen)` or flipped, right?

v1r0x:
I fixed what already mentioned in my comments as well as some other stuff to get it working on instances behind a proxy.

In some cases/machines there is a problem with `dayjs`. The plugins should be available on `dayjs` object (e.g. `dayjs.utc()`), but that's not working in some cases. Then, `dayjs.default.utc()` is required. I don't know if this is related due to switching from `require()` to `import`?
It's working on my local machine with npm 8.3.1, but not on a server with npm 8.18.0 :man_shrugging:

PR #434, opened 2023-11-23
0.10 fixes richtext

Severino:
Now the richtext should be working properly.
**Note: this requires an npm install as we now use the plain [Markdown-It](https://github.com/markdown-it/markdown-it) module for handling the markdown inputs.**


v1r0x:
I just wanted to look into enabling gfm support for the Richtext component, but ended up in a huuuuge Milkdown rabbit hole....
Long story short: On the way down I found a `editable` config in Milkdown, which allows us to use Milkdown as simple viewer without the need for another lib/plugin.

PR #439, opened 2023-12-11
[0.10] Add Entity Metadata tab

v1r0x:
- [ ] Update Changelog
- [ ] Improve _Licence_ field with predefined values
- [x] Decision what to do with _History_ section (what about Performance?)


Severino:
Richtext not filled after load.

v1r0x:
> Richtext not filled after load.

Was related to the broken Richtext attribute we already fixed. I merged `release-0.10` and now its working fine.

PR #446, opened 2024-01-26
[0.10] new profile fields

v1r0x:
Added several new fields to user profile and also reworked user info modal.


Severino (commented):
passt

PR #447, opened 2024-01-26
[0.10] Add userlist attribute

v1r0x:
Automatic notifications are missing in this branch, but I'd rather work on revamping the attribute type handling first

### DIFF
--- a/resources/js/components/EntityDetail.vue
+++ b/resources/js/components/EntityDetail.vue
@@ -852,6 +852,8 @@ export default {
                         });
                     }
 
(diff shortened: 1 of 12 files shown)


resources/js/components/EntityDetail.vue
```diff
@@ -852,6 +852,8 @@ export default {
                         });
                     }
 
+                    resetDirtyStates(grps);
```
Severino, line 855: Was fixed on other branch @v1r0x

Severino (approved):
sollte passen

PR #449, opened 2024-01-29
[0.10] Handle attribute types in separate classes

v1r0x:
Purpose of this branch/feature is to extract the parsing logic from/to the frontend and from the csv importer into separate files. Thus, adding a new attribute type should only require to add a new file in `app/AttributeTypes` and also add it as type to `AttributeBase.php`.


app/AttributeTypes/BooleanAttribute.php
```diff
@@ -0,0 +1,25 @@
...
+
+class BooleanAttribute extends AttributeBase
+{
+    protected static string $type = "boolean";
+    protected static bool $inTable = true;
+    protected static ?string $field = 'int_val';
+
+    public static function fromImport(string $data) : mixed {
+        return 
+                $data == 1 || $data == '1' ||
```
Severino, line 13: Can $data be 1? when it's always a text it will be covered by $data=='1' and inval($data)>0

Severino: $data == '1' is also covered with intval($data) > 0

app/AttributeTypes/DimensionAttribute.php
```diff
@@ -0,0 +1,35 @@
...
+use App\Exceptions\InvalidDataException;
+
+class DimensionAttribute extends AttributeBase
+{
+    protected static string $type = "dimension";
+    protected static bool $inTable = false;
+    protected static ?string $field = 'json_val';
+
+    public static function fromImport(string $data) : mixed {
+        $parts = explode(';', $data);
```
Severino, line 14: ; - see other comments

app/AttributeTypes/DropdownMultipleAttribute.php
```diff
@@ -0,0 +1,51 @@
...
+    protected static ?string $field = 'json_val';
+    protected static bool $hasSelection = true;
+
+    public static function getSelection(Attribute $a) {
+        return ThConcept::getChildren($a->thesaurus_root_url, $a->recursive);
+    }
+
+    public static function fromImport(string $data) : mixed {
+        $convValues = [];
+        $parts = explode(';', $data);
```
Severino, line 22: ; - see others

app/AttributeTypes/DropdownSingleAttribute.php
```diff
@@ -0,0 +1,36 @@
...
+    protected static bool $inTable = true;
+    protected static ?string $field = 'thesaurus_val';
+    protected static bool $hasSelection = true;
+
+    public static function getSelection(Attribute $a) {
+        return ThConcept::getChildren($a->thesaurus_root_url, $a->recursive);
+    }
+
+    public static function fromImport(string $data) : mixed {
+        $concept = ThConcept::getByString($data);
```
Severino, line 21: If I understand correctly, the import must contain the concept_url that looks like: 
https://spacialist.escience[...]/my_first_label#4631865316531653

app/AttributeTypes/EntityMultipleAttribute.php
```diff
@@ -0,0 +1,30 @@
...
+
+class EntityMultipleAttribute extends AttributeBase
+{
+    protected static string $type = "entity-mc";
+    protected static bool $inTable = true;
+    protected static ?string $field = 'json_val';
+
+    public static function fromImport(string $data) : mixed {
+        $idList = [];
+        $parts = explode(';', $data);
```
Severino, line 15: ; - see other comments

app/AttributeTypes/DateAttribute.php
```diff
@@ -0,0 +1,22 @@
...
+namespace App\AttributeTypes;
+
+class DateAttribute extends AttributeBase
+{
+    protected static string $type = "date";
+    protected static bool $inTable = true;
+    protected static ?string $field = 'dt_val';
+
+    public static function fromImport(string $data) : mixed {
+        return $data;
```
Severino, line 12: Don't we need to check for valid date here?

Severino (commented):
Found a few issues. Don't know if we must address them all in this. 
Ofteh the `;` is used as separator, but in german CSV it is regullary used in the structure so we might come up with a different separater. I proposed `|`.

app/AttributeTypes/SerialAttribute.php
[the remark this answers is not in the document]
v1r0x: Yeah, my bad. Fixed in latest commit

app/AttributeTypes/BooleanAttribute.php
v1r0x: Good point. `1` should definitely be allowed in csv. Thus I'll change type from `string` to `int|float|bool|string`

app/AttributeTypes/DateAttribute.php
v1r0x: Good point!

app/AttributeTypes/DropdownSingleAttribute.php
v1r0x: It is also possible to provide an existing label. `getByString` checks for a matching URI first and if none is found, it checks if any existing label (in any language) matches the string

app/AttributeTypes/EpochAttribute.php
```diff
@@ -0,0 +1,91 @@
...
+    protected static bool $hasSelection = true;
+
+    public static function getSelection(Attribute $a) {
+        return ThConcept::getChildren($a->thesaurus_root_url, $a->recursive);
+    }
+
+    public static function fromImport(string $data) : mixed {
+        $startLabel = 'ad';
+        $endLabel = 'ad';
+        $parts = explode(';', $data);
```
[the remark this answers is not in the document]
v1r0x: Do you have any example? I couldn't think of a real case where the provided data has a `;`. But we can look for another separator. `|` might be a problem as well. So we should think of a more complex one or allow users before uploading their data to set their own separator (with a default to `;` or `|` or ...)

app/AttributeTypes/EpochAttribute.php
```diff
@@ -0,0 +1,91 @@
...
+        if(count($parts) != 3) {
+            throw new InvalidDataException("Given data does not match this datatype's format (START;END;CONCEPT)");
+        }
+
+        $start = intval(trim($parts[0]));
+        $end = intval(trim($parts[1]));
+
+        if($end < $start) {
+            throw new InvalidDataException("Start date must not be after end data ($start, $end)");
+        }
```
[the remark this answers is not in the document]
v1r0x: Basically, yes. But since they have different count of parts and additional code, I don't know if makes sense to store the shared code into a separate function or keep this basic code as it is.

app/AttributeTypes/EpochAttribute.php
```diff
@@ -0,0 +1,91 @@
...
+        $e = $data['end'];
+        if(
+            (isset($s) && !isset($sl))
+            ||
+            (isset($e) && !isset($el))
+        ) {
+            throw new InvalidDataException(__('You have to specify if your date is BC or AD.'));
+        }
+        if(
+            ($sl == 'AD' && $el == 'BC')
```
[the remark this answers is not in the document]
v1r0x: I'd say it is a different attribute type. At least in the backend.

app/AttributeTypes/BooleanAttribute.php
Severino: Totally agree.

app/AttributeTypes/EpochAttribute.php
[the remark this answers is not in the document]
Severino: This is standard export from excel to csv with german locale.
Apprently, it wraps the value in quotes, so if the importer accounts for that,
that's not a big problem.

```csv
a;b;c;epoch
12,02;5,6;10;"20;30;10"
```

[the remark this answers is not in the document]
v1r0x: Yes, the build-in csv function in php handles this correctly

Severino (approved):
Seems good.

PR #459, opened 2024-02-15
[0.10] Several changes to core map

v1r0x:
- Fix coordinate update in popup
- Use slot to customize geodata title


app/AvailableLayer.php
```diff
@@ -113,4 +113,8 @@ public static function createFromArray($props) {
     public function entity_type() {
         return $this->belongsTo('App\EntityType');
     }
+
+    public function user_layer_option() {
```
Severino, line 117: @v1r0x is this correct in this branch? It looks like it's not used and meant for something else.

v1r0x: Yes and no :D
It is all needed for the next release of the map plugin. That's why I added it to this PR instead of adding a new one for this small feature :man_shrugging:

Severino:
@v1r0x can you have a quick look at my change. Now the overlay is moved with the point, also keeping both in focus. If that is okay you can directly merge.

Severino:
Additionally I created a PR (#501) to this branch, which adds an offset to the overlay, so that you can actually see the point. 
image

PR #463, opened 2024-02-15
[0.10] Fix for bibliography type misc mandatory 

v1r0x:


Severino (commented):
The database is requireing the title attribute to be set. So I changed the title inside misc to be mandatory, as in our case it is.

Severino:
@v1r0x please check

PR #464, opened 2024-02-20
[0.10] Add vite directive to open access blade layout

v1r0x:


v1r0x:
Closing, because #467 is a better fix

PR #468, opened 2024-03-06
Fixed wrong exported format

Severino:
Bibtex was created with the format of:
attr: {value}

but is specified as:
attr = {value}

This was fixed now.


v1r0x (approved):
:+1:

PR #473, opened 2024-03-13
[0.10] Fixes single choice

v1r0x:
See changelog for more details


Severino:
I tested this and the attribute parent is not yet properly working:
- When loading and the parent was set before. The 'child' is not containing any data to select from.
- When "saving another attribute" and then trying to save the 'child attribute' there is a 400 error that says "There is already a value set for this attribute or it is in moderation state." But this is quite inconsistent.

v1r0x:
> I tested this and the attribute parent is not yet properly working:
> 
>     * When loading and the parent was set before. The 'child' is not containing any data to select from.
> 
>     * When "saving another attribute" and then trying to save the 'child attribute' there is a 400 error that says "There is already a value set for this attribute or it is in moderation state." But this is quite inconsistent.

Both problems should be fixed in the latest commits.

PR #475, opened 2024-03-26
[0.10] Fix storing data in table & allow MC

v1r0x:


v1r0x:
Might be outdated. Due to a bug with attribute dependencies we have to rework the dirty event handling in **all** attribute types. I already started working on dependencies and event handling and it looks promising so far. But first we have to test if it works in all cases. A PR will follow.

v1r0x:
I'm closing this in favor of #479 but won't delete the branch in case the follow-up PR doesn't fix the issue.

PR #477, opened 2024-03-28
[0.10] fix misc

v1r0x:
1. After reworking how entity data is stored, the structure of returned data has changed. I forgot to update the location of the entity name in the success toast
2. Iconclass attribute had the wrong column in `$field`


Severino (approved):
I just added quotes around the entity name, as it's confusing when e.g. the name contains a whitespace.

PR #479, opened 2024-04-08
[0.10] Fix attribute dependencies & tabular attributes

v1r0x:
This succeeds PR #475


resources/js/components/attribute/Tabular.vue
```diff
@@ -117,6 +117,15 @@
                             @change="e => updateDirtyState(e, $index, column.id)"
                         />
 
+                        <date-attribute
```
Severino, line 120: Should habe been <date-range-attribute

Severino (commented):
See below

resources/js/components/attribute/Tabular.vue
Severino: Fixed in [d9d0bb2](https://github.com/DH-Center-Tuebingen/Spacialist/pull/479/commits/d9d0bb2e9ca5ba5b8f32282ca8e97cf309821f47)

PR #483, opened 2024-04-18
[0.10] Styling changes/improvements

v1r0x:


Severino:
I removed the facebook link as I am not aware of any facebook presence, otherwise we can add it again. 
From my side it would be fine to merge. @v1r0x you are welcome to check it once more.

PR #485, opened 2024-04-19
0.10 adjust import boolean x

Severino:


app/AttributeTypes/BooleanAttribute.php
```diff
@@ -2,24 +2,28 @@
...
-                strtolower($data) == 'true' || strtolower($data) == 't' ||
-                intval($data) > 0;
+    public static function fromImport(int|float|bool|string $data): mixed {
+
+        if (is_string($data)) {
+            $data = strtolower(trim($data));
+        }
+
+        return
+            $data == 1 || $data == '1' || $data == "x" ||
```
v1r0x, line 17: Non-consistent use of " and '

CHANGELOG.md
```diff
@@ -63,6 +64,7 @@ All notable changes to this project will be documented in this file.
 - Set default value in parent element in _Add Attribute_ modal
 - Values in Dropdown attributes are now sorted alphabetically
 - Values in Dropdown attributes are now searchable
+- Data Importer: Now supports the Boolean attribute with the value 'x'
```
v1r0x, line 67: I still think `x` is an ambiguous value for booleans in csv. Also it is no rocket science to replace all 'x' with '1' or something else.
But we can add it for now. For the future I'd like to have a configurable list of truthy values.

v1r0x (changes requested):

PR #488, opened 2024-04-25
0.10 fix data importer update dupes

Severino:


app/Import/ImportException.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace App\Import;
```
v1r0x, line 3: I'd move this exception to the exception namespace. We can keep the "import" subfolder, so we don't have one bloated folder.

app/Import/ImportExceptionData.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace App\Import;
```
v1r0x, line 3: I'd move this exception to the exception namespace. We can keep the "import" subfolder, so we don't have one bloated folder.

Maybe it also makes sense to move this into the `ImportException.php` file?

app/Import/ImportException.php
```diff
@@ -0,0 +1,28 @@
...
+
+        $this->data = $data;
+    }
+
+    public function getData() {
+        return $this->data;
+    }
+
+    public function getObject() {
+        info($this->message);
```
v1r0x, line 22: remove log

v1r0x (changes requested):

v1r0x (approved):
:+1:

PR #489, opened 2024-04-29
[0.10] Fix Dropdown position in entity tree

v1r0x:


v1r0x:
Is this finished?

Severino:
Yes, but as I changed the functionality quite a bit, I would like you to check it. And merge if there are no problems.

PR #491, opened 2024-05-03
[0.10] fix attribute import error

Severino:
Fixed multiple errors with the updated importer:

- Attribute errors were not displayed correctly
- Whitespaces caused imports to fail
- Fixed a display error: On small screens the control section was not scrollable, but each of the small windows were
- Fixed that name column (and parent colum) can be selected in the _Attribute-Mapping_ but the import faills without any error. Now it will import the columns despite beeing simultaneously name or parent column.


Severino:
@v1r0x can you have a look at this?

app/Http/Controllers/EntityController.php
```diff
@@ -479,27 +491,36 @@ public function importData(Request $request) {
...
+
+                    foreach ($attributesMapping as $id => $a) {
+                        if ($a == $row[$i]) {
+                            $attributeIdToColumnIdxMapping[$id] = $i;
+                            $attributeTypes[$id] = Attribute::find($id)->datatype;
+                            break;
                         }
                     }
                 }
                 continue;
```
Severino, line 515: I think here was the name column or parent column error

v1r0x (commented):
Looks good so far. I did some small adjustments

1. I excluded the `system-separator` datatype from the attribute selection list (added as new optional parameter in store getter)
2. Moved row trim function to global helper
3.1. Error message for errors during import was hardcoded in english
3.2. Error message was missing the column name (only 1-based index and datatype was shown), so I added the name to the exceptions and structs. I'm not sure if this has some side effects for any other import error. Can you please confirm there are no side effects?

app/Console/Commands/RefreshTesting.php
```diff
@@ -0,0 +1,13 @@
...
+
+    Artisan::command('test:refresh', function () {
+        $this->call('migrate:fresh', [
+            '--env' => 'testing',
+        ]);
+        $this->call('db:seed', [
+            '--env' => 'testing',
+            '--class' => 'DemoSeeder',
+        ]);
+    })->describe('Resets the testing database and seeds the demo data.');
```
v1r0x, line 13: I'm not sure if we should "polute" the test namespace with our own commands. Maybe create our own spacialist namespace where we can also add other commands (e.g. create users from cli, ...)

v1r0x (changes requested):
Beside the artisan command: :+1:

PR #493, opened 2024-05-07
[0.10] Several BibTex fixes

v1r0x:
This PR is still missing a fix for the broken file picker on mac (not able to select any bibtex-related files) and windows (able to change file type to "any"). @Severino could you please check this at least for windows? Might be the same fix for mac.


v1r0x:
I added another fix, because adding new items of type `inbook` was not possible due to broken validation (no fields displayed)

Severino:
I forgot to add the fix for the files that were not selectable on windows and map. Those are now in the PR #500

PR #495, opened 2024-05-10
[0.10] Fix logging seeders

v1r0x:


Severino:
Looks good. I verified there are no activities in the db after initial seeding.

PR #498, opened 2024-05-13
Change styling slightly to be more clear

Severino:
I propose small style changes to the certainty style:

image

+ Make Comment and Bookmark less prominent by usnig outline icons
+ Decrease visibility of the unset certainty, to be less confusing when seeing question marks everywhere
+ Changed the exclamation icon to a question mark, as it is (at least in all my projects so far) used to mark uncertainty (e.g. in spreadsheets)
+ Increase the sup size sllightly: 0.75rem -> 0.8rem


Severino:
I'm also not sure about the 'blue' color. 
IMO the following would be more suitable.
red > orange > yellow > green (only 100%)

We would need to introduce those variables for a theme (at least the orange).

v1r0x:
> * Make Comment and Bookmark less prominent by usnig outline icons

For the sake of consistency I'd go with the solid icons. FA has 10 times more icons in the solid category than in the regular.

> * Decrease visibility of the unset certainty, to be less confusing when seeing question marks everywhere

I find it very hard to see. Either we increase the opacity or keep it as it is. Or use another (light) color e.g. `text-body-tertiary`

> * Changed the exclamation icon to a question mark, as it is (at least in all my projects so far) used to mark uncertainty (e.g. in spreadsheets)

I like the idea of having different icons for unset (?) and set, but uncertain (!) and of course certain (:heavy_check_mark:)

> * Increase the sup size sllightly: 0.75rem -> 0.8rem

In solid style they are quite big then.

v1r0x:
> I'm also not sure about the 'blue' color.
> IMO the following would be more suitable.
> red > orange > yellow > green (only 100%)

I agree, the teal color is not the best fit. But I'd tackle this in the next release(s) when we (might) introduce a couple more themes and/or work on a separate component package

PR #499, opened 2024-05-16
[0.10] Added processors to transform type value and all tags to lower-case

Severino:
I noticed by accident, that the importer fails when trying to load a file that is not all lower case (for type and keys).
Bibtex is like always totallly chill with everything that you throw at it (thanks for nothing 🙄).

So I added two processors to the convenient BibTex parser to:

1. Transform all tag keys to lower case: AUTHOR -> author
2. Transform all type values to lower case: @Article or @ARTICLE -> 'type': 'article'


v1r0x:
Is this still needed or is it already part of #500 ?

Severino:
Yes it's still needed. No, it's not part of 500. I opened a separate PR for this. 😉

v1r0x:
From #500 

grafik
grafik

Severino:
You are correct. Maybe I errorously created the fix-mac-win branch from this branch.
But then it's already merged.

PR #500, opened 2024-05-16
0.10 fix bibtex files non selectable on win mac

Severino:
Fixing the file explorer not showing .bib and .bibtex files on windows.
Somehow the 'extension' prop of the file-upload component was not applied.
So I added it to the accept prop (which shold be internally merged anyways, as it is the
only available attribute on the input element where it has an effect).

Tested on Windows and MacOs.


v1r0x (approved):
Looks good

PR #502, opened 2024-05-17
0.10 feature si attribute rework

Severino:
I'm currently still working on this.
I reworked the commit to be more in line with
what we discussed.

But the nature of PHP only allowing Classes to be 
imported using the 'use' keyword made me fallback to the
UnitSystem classes.

I commented the approach in more details
in the "Files changed" tab.


app/AttributeTypes/SiUnitAttribute.php
```diff
@@ -2,338 +2,27 @@
...
-            ],
-            // 'substance' => [],
-            // 'luminous' => [],
-            // 'energy' => [],
-            // 'discspace' => [
-            //     // b, kb, mb, gb, tb, pb
-            //     // bi, kibi, mibi, gibi, tibi, pibi
-            // ],
-        ];
-    }
```
Severino, line 268: We both didnt like the verbosity of the config array here, so I tried to move this to their respective classes.
I could not get it to work recreating the simple array structure so I sticked with the "UnitSystem" approach.

app/AttributeTypes/SiUnitAttribute.php
```diff
@@ -2,338 +2,27 @@
...
-                } else {
-                    if($unit[$key] == $value) {
-                        return $unit;
-                    }
-                }
-            }
+    public static function getGlobalData() : array {
+        $unitsArray = [];
+        foreach(UnitManager::get()->getUnitSystems() as $unitSystem) {
+            $unitsArray[$unitSystem->name] = $unitSystem->toArray();
```
Severino, line 22: Renamed from 'addGlobalData' as we just retrieve the unit system array. 
Not adding anything.

app/AttributeTypes/SiUnitAttribute.php
```diff
@@ -2,338 +2,27 @@
...
-                if(is_array($unit[$key])) {
-                    if(in_array($value, $unit[$key])) {
-                        return $unit;
-                    }
-                } else {
-                    if($unit[$key] == $value) {
-                        return $unit;
-                    }
-                }
-            }
```
Severino, line 284: Moved to "UnitManager" and "UnitSystem"

app/AttributeTypes/SiUnitAttribute.php
```diff
@@ -2,338 +2,27 @@
...
-        }
-
-        $unit = self::getUnitFromSymbol($value);
-
-        if(isset($unit)) {
-            return $unit;
-        }
-
-        return null;
-    }
```
Severino, line 311: Moved to "UnitManager" and "UnitSystem"

app/AttributeTypes/SiUnitAttribute.php
```diff
@@ -2,338 +2,27 @@
...
-
-    private static function normalizeValue(mixed $value, array|string $unit) : mixed {
-        if(!is_array($unit)) {
-            $unit = self::getUnitFromUnknown($unit);
-        }
-
-        return $unit['convert']($value);
-    }
```
Severino, line 319: Moved to "Unit"

app/AttributeTypes/Units/Call.php
Severino, line 1: The "Unit" requires callables that are most often based on the same functions. Therefore they are collected in here.
I used it "Callables" in the beginning but thought "Call::multiple" feels more natural.

app/AttributeTypes/Units/Constants/Imperial.php
Severino, line 1: Struggled with this one, as the constants need a special entry in the `composer.json` and require a `composer dump-autoload`

app/AttributeTypes/Units/Implementations/AreaUnits.php
Severino, line 1: With the "UnitSystems" I would have preferred the approach we talked about. And I tried to do it in the beginning to just have the 'arrays' separated into their systems using the "UnitSystem" class. But could not do that in PHP:

@areasystem.php
$AreaSystem = new UnitSystem(name, baseUnit)

@siunitattribute.php

use App/AreaSystem

$groups = [
        'area' => $AreaSystem
]

app/AttributeTypes/Units/SiPrefix/Si.php
Severino, line 1: I struggled with the names for the now called 
- Si.php -> All implementations of the SiPrefixes
- SiPrefix.php -> BaseClass for the SiPrefixs

In the end I thought Si::$KILO is kinda easy to write.

I also struggled witht he export functionality in PHP. As I wanted to have the autocompletion for all the values. But as I wanted to have them as the SiPrefix class, const were no option. Therefore this is now a self initializing class. I don't know if that is good or not. But it's working as expected. The static class would be an alternative for the 'Imperal.php' constant exports which I really do not like, as it requires the autoloader dump.

app/AttributeTypes/Units/SiPrefix/SiPrefix.php
Severino, line 1: I just like this, it does what it should. When thinking about the naming of SI and SiPrefix, I thought this class could be renamed to something more general, as it is just a label, symbol and a value. But I didn't want to spend too much time on it so it's just named the same.

app/AttributeTypes/Units/Unit/BaseUnit.php
Severino, line 1: I sticked with the "BaseUnit" as I think it makes a lot of sense as a concept. **HOWEVER** now the unit is the ParentClass and the "Base Unit" is the subclass. I think this is great, as then for a UnitSystem we must pass a BaseUnit, but we cannot pass a Unit as it is not a BaseUnit.

app/AttributeTypes/Units/Unit/SiUnit.php
Severino, line 1: We talked about the possibility to use the SiPrefix to create a Unit. And my first idea was a separate class that takes a Prefix and a BaseUnit it returns a new Unit. But as the concept of Units already existed I thought a SiUnit class is more comprehensible. So that was what I went for.

app/AttributeTypes/Units/Unit/Unit.php
Severino, line 1: I kept the Unit as is, it's now the superclass to the BaseUnit (see BaseUnit for more info).

The 'convert' was renamed to 'normalize' as you used it before, which is simply more accurate. And I added the "toObject" method that is used by the "UnitSystem" to create an Array representation.

app/AttributeTypes/Units/UnitSystem.php
Severino, line 1: Stayed mostly the same. Added some getter methods. An 'addMultiple' method, which makes it much cleaner when implementing a UnitSystem. And the toArray() method.

app/Http/Controllers/HomeController.php
Severino, line 1: Just renaming, as there is no data added:
addGlobalData=> getGlobalData
addData => datatypeData

app/Models/Singleton.php
Severino, line 1: I used the singleton pattern inside the "UnitManager" so I created this reusable base class in the Folder "App\Models" if you have better ideas for the location, we can adjust this.

app/Providers/AppServiceProvider.php
```diff
@@ -96,26 +96,26 @@ public function boot()
...
             $refField = request()->input($parameters[0]);
-            $baseunits = SiUnitAttribute::getUnits();
-            if(!in_array($refField, array_keys($baseunits))) {
-                return false;
-            }
-            $units = $baseunits[$refField]['units'];
-            foreach($units as $unit) {
-                if($unit['label'] == $value) {
-                    return true;
-                }
```
Severino, line 114: Moved that functionality to the UnitManager

app/Providers/AppServiceProvider.php
```diff
@@ -96,26 +96,26 @@ public function boot()
             return in_array($value, array_keys(Bibliography::bibtexTypes));
         });
         Validator::extend('si_baseunit', function ($attribute, $value, $parameters, $validator) {
-            return in_array($value, array_keys(SiUnitAttribute::getUnits()));
+            return UnitManager::get()->hasQuantity($value);
```
Severino, line 99: Moved that functionality to the 'UnitManager'. I'm unsure about the naming. In English it's called the 'quality' in which 'domain' the value resided e.g. length or area. But otherwise it's just the name of the UnitSystem in our context. So that's debatable.

composer.json
Severino, line 1: Changes due the cont exports in the imperal.php

tests/Unit/Attributes/SiAttribute.php
Severino, line 1: I still need to fix all the si unit tests

Severino (commented):
I left some explanations about my reasoning of this refaactoring job.

v1r0x:
I tried to simplify the unit structure/logic and think this is a good tradeoff between readability/consistency and abstraction/oop.
I also added some minor fixes for different warnings (wrong component prop types, i18n) and moved the singleton class from (former) models folder to patterns folder.

PR #504, opened 2024-06-06
[0.10] Fix table performance & rework attribute(list) logic

v1r0x:
This branch started as a simple performance booster for table attributes. But the whole attribute handling was a huge code block with DRY code in the table component as well.
Thus, I reworked the attributelist and table component to be based on a new `Attribute` component (tables also has a new `TabularRow` component). This introduced a new problem with entity attributes (especially in tables), so I tried to move all logic that loads the data needed for the frontend to the `serialize` and `unserialize` in the AttributeType classes.


resources/js/components/attribute/Attribute.vue
```diff
@@ -0,0 +1,428 @@
...
+            } = toRefs(props);
+            // FETCH
+
+            const attrRef = ref({});
+            const state = reactive({
+                type: computed(_ => data.value.datatype),
+                disabled: computed(_ => data.value.isDisabled || disabled.value),
+                value: computed(_ => getValueOrDefault()),
+                // TODO check for selection need?
+                selection: computed(_ => getAttributeSelection(data.value.id) || {}),
```
Severino, line 359: This is a problem, I think this should return an array as default => ... | [ ]
Throws error when creating table attribute.

resources/js/components/attribute/Attribute.vue
```diff
@@ -0,0 +1,428 @@
...
+        emits: ['change', 'update-selection'],
+        setup(props, context) {
+            const {
+                data,
+                valueWrapper,
+                disabled,
+                reactTo,
+                hideLinks,
+                preview,
+                previewData,
```
Severino, line 349: Those are not used.

resources/js/components/attribute/Tabular.vue
```diff
@@ -620,8 +510,8 @@
                 initialValue: value.value,
             });
             const state = reactive({
-                isPreview: computed(_ => previewColumns.value && Object.keys(previewColumns.value).length > 0),
-                columns: computed(_ => state.isPreview ? previewColumns.value : getAttribute(attribute.value.id).columns),
+                isPreview: computed(_ => preview.value && previewData.value && Object.keys(previewData.value).length > 0),
```
Severino, line 513: Why isn't preview.value check enough?
Maybe a different name if it does more than one thing?

resources/js/components/attribute/Tabular.vue
```diff
@@ -35,183 +36,37 @@
...
-                            :value="row[column.id]"
-                            @change="e => updateDirtyState(e, $index, column.id)"
-                        />
-                    </td>
-                    <td
-                        v-if="!disabled"
-                        class="text-center"
+                        v-if="row.hidden_info"
+                        class="text-muted text-center fs-5 p-2 bg-primary-subtle"
+                        :colspan="Object.keys(state.columns).length + 2"
```
Severino, line 46: use computed at this and the next position? (redundancy)

resources/js/helpers/helpers.js
```diff
@@ -399,6 +403,48 @@ export function getInitialAttributeValue(attribute) {
...
+
+    switch(type) {
+        case 'boolean':
+            return false;
+        case 'geography':
+        case 'iconclass':
+        case 'string-sc':
+        case 'date':
+        case 'url':
+            return '';
```
Severino, line 417: There are 3 positions where we return ' '. Combine the non-default cases that return ''?

Severino (commented):
In current state the creation of a table attribute is not working.

resources/js/components/attribute/Attribute.vue
v1r0x: These are some remains from stuff I copy&wasted 😅 can be deleted 👍

resources/js/components/attribute/Attribute.vue
v1r0x: Yeah, selections should be Arrays 👍

resources/js/components/attribute/Tabular.vue
v1r0x: 👍

resources/js/helpers/helpers.js
v1r0x: Yes, we can add this at the end right before `default`. Thought I already did that 🤔

resources/js/components/attribute/Tabular.vue
v1r0x: I just did a double check to make sure that we enabled preview (`preview.value`) and have actual data to preview (`previewData.value`).

v1r0x: I think there was a mix up from the old structure. `preview` and `previewData` is part of `Attribute.vue` and `Tabular.vue` only receives `previewColumns`. Thus, checking for `previewColumns` existance and content is enough.

v1r0x:
I resolved merge conflicts and added our new si-unit-attribute, that was added to the pre-PR attribute handling logic. Can you also confirm it is still working @Severino ?

Severino:
I tried getting it to work, but when I create a new and empty table with an SiUnit Attribute, I cannot add a new line, as it creates an empty object, however the si attribute need a metadata object for the bast_unit and the default_unit

PR #506, opened 2024-06-17
0.10 multi move

Severino:


Severino:
This was abandoned and revisited with #600

PR #525, opened 2024-10-29
10.1 dm transfer

Severino:
Exports the datamodel of elements in the entity hierarchy.

Needs a transfer connection inside the env file:

```
DB_TRANSFER_CONNECTION=pgsql
DB_TRANSFER_HOST=localhost
DB_TRANSFER_PORT=5432
DB_TRANSFER_DATABASE=target_db
DB_TRANSFER_USERNAME=user
DB_TRANSFER_PASSWORD=user
```

Can be run with:

`php artisan app:export --entity=31`


app/Console/Commands/ExportEntityTypes.php
```diff
@@ -0,0 +1,563 @@
...
+            // This path did not return anything. I return 0 here (see bash return codes)
+            return 0;
+        }
+    }
+
+    private function extractThesaurusConcepts($model, array &$allConcepts): void {
+        $collection = ThConcept::where('concept_url', $model->thesaurus_url)->get();
+        foreach($collection as $concept) {
+            $allConcepts[$model->thesaurus_url] = $concept;
+        }
```
v1r0x, line 128: Does it make sense to iterate over an array here?

1. There should be only `concept_url` matching the model's `thesaurus_url` in the DB (`concept_url` is unique in `th_concept`-table)
2. Looping over `collection` and setting `allConcepts[$model->thesaurus_url]` will only set the content of the last item or am I missing something?

app/Console/Commands/ExportEntityTypes.php
```diff
@@ -0,0 +1,563 @@
...
+        return $collection->labels->first()->label;
+    }
+
+    private function collectAllDataFromEntityChildren(Entity $entity): array {
+        $thesaurusLabels = [];
+        $entityTypeIds = $this->getEntityTypes($entity);
+        $entityTypes = array_map(function($entityType) {
+            $entityType = EntityType::findOrFail($entityType);
+            return $entityType;
+        }, $entityTypeIds);
```
v1r0x, line 142: Simply return `EntityType::...` should be enough. Also eliminates confusion with duplicate variable name

app/Console/Commands/ExportEntityTypes.php
```diff
@@ -0,0 +1,563 @@
...
+            foreach($entityType->attributes as $attribute) {
+                $fetchedAttribute = Attribute::findOrFail($attribute->id);
+
+                if($attribute->thesaurus_root_url) {
+                    $rootConcept = ThConcept::where('concept_url', $attribute->thesaurus_root_url)->first();
+                    $thesaurusConcepts[$attribute->thesaurus_root_url] = $rootConcept;
+                    $thesaurusRootConcepts[$attribute->thesaurus_root_url] = $rootConcept;
+                }
+
+                if($attribute->id === 1) {
```
v1r0x, line 160: What's the purpose of this check?

app/Console/Commands/ExportEntityTypes.php
```diff
@@ -0,0 +1,563 @@
...
+        $this->printSeparator();
+        $this->printHeading("Transfering Attributes", $data['attributes']);
+        $this->transferAttributes($user, $data['attributes']);
+        $this->printSeparator();
+        $this->printHeading("Transfering Entity Type Attributes Relation", $data['entityTypeAttributeRelations']);
+        $this->transferEntityTypeAttributeRelations($user, $data['entityTypeAttributeRelations']);
+        $this->printSeparator();
+    }
+
+    private function selectOrCreateTransferUser(): User {
```
v1r0x, line 293: We shouldn't create a dummy user. This is going to be a mess if data is stored in a repository.

PR #528, opened 2024-10-31
0.11 feature entity tree export

Severino:
Adds an export function to the Entity tree's context menu.
This will export the selected item and all it's children. 

This will create either a csv file if all entities are of the same type. 
Or a zip file containing one csv file each for every entity type.
This export may be used to transfer data from one spacialist instance to another.

Note: This does not export any data models or thesaurus concepts.


v1r0x (approved):
:+1: :penguin:

PR #531, opened 2024-11-13
[0.10.1] Fix integer validation

v1r0x:


app/Http/Controllers/EntityController.php
```diff
@@ -709,36 +709,44 @@ function setOrUpdateImportedAttributes($entity_id, $row, $headerRow, $attributeI
...
         }
 
         $addedAttributes = [];
-        foreach ($request->request as $patch) {
+
+        if(count($request->request) === 0) {
+            return response()->json([
+                'entity' => $entity,
+                'added_attributes' => $addedAttributes,
+            ]);
```
Severino, line 732: Shall we pass a 204 to inform the client that no data was passed? Maybe good for future debugging purposes.

resources/js/components/attribute/Integer.vue
```diff
@@ -84,11 +88,19 @@
...
+                if(!validateValidation(v.value)) return;
+                // also call validate() to make sure all values are up to date
+                // without this, meta/errors still contain the value from last validation
+                validate().then(validation => {
+                    context.emit('change', {
+                        dirty: v.meta.dirty,
+                        valid: v.meta.valid,
+                        value: v.value,
+                    });
+                })
```
Severino, line 103: The value is correctly validated. But:

1. The value being invalid does not prevent the user from submitting => User inserts invalid value and has no indicator to see her mistake => submits (is not prevented) => sees data as if it was updated successfully => data infact remaind the same.
2. Should we just block users from entering other values than numbers (and leading - ?)

Severino (changes requested):
Invalid input value should prevent user from submit.

Severino (approved):

PR #532, opened 2024-11-13
[0.11] Display path of thesaurus search results

v1r0x:


Severino:
This should work now appropriately. Please check again. Also adjusted the test case.

PR #533, opened 2024-11-14
[0.10.1] Add Tab action to Single Choice dropdown attribute

v1r0x:


Severino:
I added a badge to provide some feedback on what can be 'tabbed' and also the ability to clear the text and selection by pressing the 'delte' key.

If you approve of that I will push my changes.

tab single choice

PR #534, opened 2024-11-15
[0.11] Extend entity contextmenu with options to add above/below

v1r0x:
Requires 0.11-feat-realtime (#516) to be merged first


Severino:
This was implemented using a cherry-pick #554

PR #540, opened 2024-12-05
[0.11] Move files to private storage

v1r0x:
This PR requires #516 because the authentication is updated from JWT to a Cookie-based approach. This is essential for this PR.


v1r0x:
Closed in favor of #557

PR #541, opened 2024-12-05
[0.11] Handle colors on entity types

v1r0x:


Severino:
@v1r0x for me this is looking good.

PR #546, opened 2024-12-11
Fixed importer when header is not present and row numbers are used.

Severino:
I tried to import a CSV without a header. And noticed that it fails.
This pull request fixes that and adds all tests for the 'NoHeaders' variant.


v1r0x:
We can ignore the failing tests. This is due to a change in github's workflow system. `ubuntu-latest` now refers to 24.04 and the required ppa has not (yet) packages for this release. A fix is already in #547, so we can re-check tests in release-0.11 before release

PR #547, opened 2024-12-13
[0.11] Add dependency groups to attributes

v1r0x:


Severino:
Problems I found (I only worked with booleans so far):

- when saved in wrong state, the dependent objects wont show up
- Cannot save dependency when 'entspricht = false'

PR #548, opened 2024-12-13
[0.11] feat realtime review sev

Severino:
- Reworked comment section in store from handleComment => addComment,updateComment and deleteComment respectively. Functions should only have one task and this is more readable.
- Fixed a problem with the created comments being send to the creator himself, resulting in a duplicated entries
- Reworked the activeUserIndicator in style and excluded the current user, as the current user should be aware he is on that site


Severino:
activeUserIndicator

v1r0x:
My opinion regarding the active user list:
- I'd keep the current user, but we can hide the whole list if they are the only user
- I'd also keep the **green** dot indicator, because it is the default indicator for _online_ in (almost) every other website/app/program
- The border breaks with the rest of the row layout and draws focus. I tried a more subtle design with an almost transparent background and no border, but IMO no border and no background looks best and matches the other parts
- The blinking dot is also distracting and draws focus

active-users

Severino:
Having often problems with windows that are not connected, I added a component to provide the user with feedback when they are ....

not connected:
indicator-no-connection

and if reconnected 
indicator-reconnected
(vanishing after 3s):

v1r0x:
grafik

We already have a Toast system. I think it would be better to have one system for updates/information.

Severino:
I have now fixed all attributes (except for the table) to properly update when another user updates the attribute value.
But there are still errors happening, that I would pass back to you @v1r0x because I'm not so deep into
the system to quickly fix it.

Some properties do not update showing the correct blue dot, but the red dot. 
(Somehow they are set to dirty after the value was updated). The 
affected attributes are:

- [ ] Dimensions
- [ ] User List
- [ ] Entity MC
- [ ] String MC
- [ ] Time Span
- [ ] Time Span & Epoch
- [ ] Si Unit

The Si Unit has further problems, as it does not update on the other side.
After a save it's sets dirty. But checking inside the watcher if the changed values are
identical always have the same value in the old/new arguments.

Severino:
I moved the connection error to the toaster.

v1r0x:
All attributes (except table) should update fine now. New table rows are correctly updated over websockets. But updating existing rows/columns does not work. Adding a watcher on `_ => value.value` instead of `_ => value` fixes this problem, but has to be done in all attributes.
Beside that, there is a problem (unrelated to this PR) that most attributes don't store the updated value after save. Thus, resetting an already saved entity results in wrong values (e.g. change int field from 1 to 2 -> save -> change from 2 to 3 -> reset; results in 1 instead of 2).
There seems something to be really wrong with updating values. We should handle this in a follow-up PR. I think this is related to table rows not updated over websockets. I'd suggest to merge this branch (if you don't encounter any other errors)  and then focusing on fixing attribute updates in another PR.

PR #549, opened 2025-02-03
{0.13] ref list attribute style

Severino:
The List attribute's style is still quite rough.

I proposed a restyle which will be posponed and further evaluated for version 0.13

The new style looks initially like this:
image


v1r0x:
Is now part of dhc-components

PR #553, opened 2025-02-07
[0.11] Refactor geodata-package

v1r0x:
Is this branch finished or do we need to fix anything?


v1r0x:
After discussion with @Severino we agreed on merging

PR #554, opened 2025-02-11
[0.11] CHERRY add option to add entity above/below to entity tree contextmenu

Severino:


Severino:
#534 was in a bad state. So it was cherrypicked to this branch,

Severino:
closes #534

PR #558, opened 2025-02-12
[0.11] feat private files cp review sev

Severino:
Added a FileDirectory class that should handle all files inside one directory.
And prevents foreign files to be tempered with.

### DIFF
--- a/app/Bibliography.php
+++ b/app/Bibliography.php
@@ -2,8 +2,8 @@
 
 namespace App;
 
+use App\File\FileDirectory;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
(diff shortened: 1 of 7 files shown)


v1r0x (changes requested):
:+1: from me

But:
- I changed names and parameters at different places
  - e.g. `disk` should be second parameter in (File)Directory constructor and default to `'local'`
  - reworked `validate` method to also check for existence and renamed it to `contains` as it is IMO clearer what it does
- I added missing return types (and found that sometimes a wrong type was used/returned)
- Removed `delete` method because it was never used. Renamed `deleteRaw` to `delete`, because it simply deletes the file. So why is it "raw"?
- I removed the try/catch in `BibliographyController` (see code comment), because no method throws an exception. Please check if I missed something

PR #559, opened 2025-02-13
[0.11] feat entity references cp review sev

Severino:
In this PR I reworked the proposed entity-reference in some various aspects due to various problems;
- Entity Reference did never show
- Submit did not wait for request (User could e.g. double click to create 2 instances) when creating new references
- Api was used when only store should access the api
- Updates were not applied appropriately
- Redundance in the code due to multiple places we need the reference components.

___

What was changed:
- Created a _ReferenceTab_ component that massively reduced and streamlined code inside _Main_ (can use <component :is=?/>
- Created a _QuotationInput_ component for creating new Quotations.
- Created an _EditableQuotation_ component that can be toggled to edit between Quoatation and QuatationInput. Massively simplifies the 'edit' logic from _Reference_ and allowed to easily add the logic to the EntityReferences in the _ReferenceTab_ which was not yet possible.
- Fixed a problem in the _Citation_, that titles were not formatted properly.
- _QuotationInput_ and _ReferenceForm_ now do have a 'pending' state that disables the submission when request was already sent.
- Emoji button caused an error. Refactored to use ref instead of 'getElementById'
- Improved the styling of _ReferenceForm_ which was really weird. For entity refs, it is now fixed at the bottom of the tab page (literature dropdown opens to top), see image.

entity_ref


Severino:
Should be ready to be reviewed. My tests worked for both local and websockets now. 👍

v1r0x:
The reference handling in store was kinda weird
- duplicate code in entity store and (new) reference store (I removed and replaced the references of the code from entity store)
- missing variables in reference store

I also changed several parts:
- Got rid of AutoTextarea component (IMO it is weird to have this with no option to resize or stop growing/shrinking)
- Renamed QuotationInput to QuotationForm as it is not only an input field, but also buttons
- Moved QuotationForm buttons to new line and added labels
- Added (empty) reference store keys to match all the other stores
- Fixed wrong `type` reference in Bibliography Details component (has to be `data.entry_type`)
- Restore old `userRequest` method as the parameter was never used and a lot more tests failed

Severino:
I would rebase merge this onto the original branch and then rebase merge this into the release branch via interface.
The only conflict on the original branch is the changelog

v1r0x (approved):
:+1:  from me

PR #566, opened 2025-03-12
[0.11] Add dependency groups to attributes

Severino:
This branch was squased and rebased from #547 
There were some further changes:

- Renamed 'union' / 'intersection' to 'or' / 'and' (the old was for me very confusing)
- Restructured a lot of the code into components.
- 'Mode Toggle' is now flipping the different 
- Added number composable to number inputs (that I could test more easily)
- The dependency form now uses the appropriate <Attribute> as input element.
- Checked that it works with _System Separators_  (already implemented)


v1r0x:
Still working with all changes and renaming :+1:  Tests also look good. Merge? @Severino

v1r0x (approved):
:+1: :penguin:

PR #569, opened 2025-03-26
[0.11.1] Make attribute lists striped

v1r0x:
This PR makes attribute lists striped to better distinguish them from each other (especially from tables).
Tables now needed a neutral bg, because it uses the same "stripe color". I also removed table borders to save some space (and they are also already striped).

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 All notable changes to this project will be documented in this file.
 
 ## 0.11.1
+### Added
+- Attribute Lists are now striped for better readability
 ### Fixed
 - Login not possible on instances in subfolder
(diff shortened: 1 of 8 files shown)


Severino:
I mentioned to you (a while ago) that I don't like this change. It demolishes the clean look of the entities while providing little to no benefits:

The reason for striped lists is for the eye to not lose track of the row in focus. However, in the EntiyDetails the labels are right next to a single input element which leaves not much room for losing track of said row. Therefore I would just leave it as is.

v1r0x:
In some cases it is hard to distinguish attributes and their labels (e.g. with tables and several attributes of the same type). But I'm fine with keeping the current style.

PR #572, opened 2025-04-30
[0.12] Add 2FA

v1r0x:


Severino:
@v1r0x I don't like the 2FA just being losely in the Profile section. 
It feels out of place.  

I would rather add it into the settings, here the user may be also able to change their password*:
image

What do you think? 

* for another PR

Severino:
I also found a severe security vulnerability with the current implementation, that makes the user secret public to the user himself and also all other logged in users. It looks like the fortify trait on the user don't hide those fields by default and sends them when the models are fetched. I assume adding this to the User should fix the issue:
```
    protected $hidden = [
        'password', 'remember_token', 'two_factor_secret', 'two_factor_recovery_codes',
    ];
    ```

v1r0x:
I think in the long run it makes sense to merge the profile and settings into one site. We could keep the _Profile_ item in the user dropdown, but redirect to the user preferences start page.

The 2FA in the settings also feels out of place and doesn't look that nice. We should think of a (minimal) redesign of the preferences pages. But IMO it has a very low priority for now.

Severino:
Okay then we postpone this for a future preferences rework.

PR #573, opened 2025-04-30
0.11 fix pre loading time

Severino:
I did this as a local fix, for a faster loading time.
It moves all data that is required by the initialize app
directly into the pre call. Only requiring a single request
for the initialization (without this data, the prigram will go in error mode anyways).

Would avoid scenarios like "Cannot read tags" as the permission is no longer checked.

In my DB without this it takes 1995 ms. With this fix it takes 615ms.

### DIFF
--- a/app/Attribute.php
+++ b/app/Attribute.php
@@ -84,6 +84,30 @@ class Attribute extends Model
         }
     }
 
+    public static function getSelectionsFor(&$attributes): array {
+        $selections = [];
+        foreach($attributes as $a) {
+            $selection = $a->getSelection();
+            if(isset($selection)) {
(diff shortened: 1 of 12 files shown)


v1r0x:
I already started a local update on this branch. I don't like to add most of the logic to the HomeController. So I moved it to existing models and added a global "model" for e.g. version data logic. We should also think about the - now - obsolete api endpoints on js and php side. Should we keep them, because we _may_ need them in the future again? Or should we remove them, because they are no longer needed and everything is handled in the `pre` endpoint?

v1r0x:
I pushed my changes and removed all unused api functions. Please review and feel free to merge.

PR #575, opened 2025-05-30
[0.11.1] Several Fixes

v1r0x:


resources/js/bootstrap/plugins.js
```diff
@@ -120,11 +120,13 @@ export const SpPS = {
         plugins: {},
         app: null,
         t: null,
+        pinia: null,
     },
-    initialize: (app, t) => {
+    initialize: (app, pinia, t) => {
```
v1r0x, line 125: Is it not possible to import pinia in plugins.js?

Severino: I think it's good to pass it after initialization and it's more clear what the main.js set's on the SpPS. But as it's an object it should make no functional difference when accessing it before or after the initialization.

app/Utils/UrlUtils.php
```diff
@@ -9,13 +9,12 @@ class UrlUtils {
...
+    static function getSubPath($url) : string {
         $matches = [];
-        $doesMatch = preg_match('/https?:\/\/.+?\/(.*)/', config('app.url'), $matches);
-        
+        $doesMatch = preg_match('/https?:\/\/.+?\/(.*)/', $url, $matches);
         if(!$doesMatch || !isset($matches[1]) || empty(trim($matches[1]))) {
             return '/';
         }
-        return trim($matches[1]);
+        return '/' . ltrim(trim($matches[1]), '/'); // remove leading and trailing slashes
```
v1r0x: Does it remove trailing slashes?
You can use `Str::start()` and `Str::finish()` instead of `/` + `ltrim()`.

PR #579, opened 2025-09-01
[0.12] Allow plugins to register attribute types

Severino:
I rebased the 0.12 branch and also the 0.12-attribute-plugin, and fixed some problems:

+ AttributeBase: 
  + only return label if it exists
  + getHasSelection returns bool now (string before)
+ HomeController used `App\AttributeTypes\AttributeRegistry` instead of `App\Registries\AttributeRegistry`
+ Plugin.php
  + Removed `getWithMetadata` return type array (produced errors)
  + Fixed store to use `$publishScript` instead of `$filehandle` in `directory->store`
+ Attribute Registry
  + Fixed only using php basename to create plugin class (didn't work, required full namespace path)
  + Added safeguards and logs, when plugin namespace cannot be found or baseclass is not of AttributeType.
+ MainView.vue: Prioritize component over componentTag 
+ Attribute Template: made attribute sorting cleaner taking attribute labels into account.


tests/TestCase.php
```diff
@@ -28,10 +28,6 @@ abstract class TestCase extends BaseTestCase {
      */
     protected $seeder = TestingSeeder::class;
 
-    protected $connectionsToTransact = [
-        'testing'
-    ];
-
```
Severino, line 34: This somehow prevented the RefreshDatabase to work properly. 
I think I added it at some point as the RefreshDatabase did not work properly as it was mentioned here:
https://github.com/laravel/framework/issues/51674#issuecomment-2146532360

.env.phpunit
Severino, line 1: I removed this as it seems redundant, as the phpunit.xml and the .env.testing should already take care of the testing environment. Testing did work more reliably after removing this.

PR #580, opened 2025-09-02
Removed reduntant api calls and improved metadata handling

Severino:
I noticed the application being very slow when changing entities.
This was due to the endpoints being called twice for every entity change and
additionally due to a lot of heavy lifting being done in every middleware on every request.

In this PR I removed the redundant api calls, fixed the metadata tab* only fetching the data when being visited and combined all data that needs to be retrieved
by the entity detail in a single endpoint:
GET::v1/entity/{id}/entity_detail

That returns the data:
```php
[
         'data' => $data,
            'metadata' => $entity->getAllMetadata(),
            'references' => Reference::getByEntity($id),
            'parentIds' => $entity->parentIds,
            'parentNames' => $entity->parentNames,
            'attributeLinks' => $entity->attributeLinks,
]
```

This results in way better performance and therefore a much better responsiveness of the app.

* Furthermore when adjusting the unecessary request with the metadata tab I fixed an error throwing an error when saving the metadata tab (unknown variable).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,14 @@
 All notable changes to this project will be documented in this file.
 
 ## 0.12 - (Luxor)
-### Add
-- New Api endpoint for getting all entity details data in one request: GET::v1/entity/{id}/entity_detail
-
+### Added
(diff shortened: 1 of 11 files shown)


v1r0x (approved):
:+1:

PR #581, opened 2025-09-10
Fixed ImageTIle and layer error

Severino:
The spacialist map.js caused errors when using the map plugin.

  - Fixed using undeclared & deprecated _TileImage_, now using _ImageTile_
  - Fixed using undefined _layer_ in map.js, now using _layerdata.layer_


v1r0x:
was already fixed in https://github.com/DH-Center-Tuebingen/Spacialist/tree/release-0.12 
The rebase removed half of it. We should rethink that approach.

PR #582, opened 2025-09-10
[0.12] Added filesystem migrations

Severino:
Enhanced the migration system with project-specific base classes and filesystem migration support

- Added `App\Migration\Migration.php` base class that disables logging by default for all subclasses
- Created custom `make:migration` command that generates migrations by extending the project-specific base class
- Introduced `App\Migration\FilesystemMigration` class for migrations affecting the filesystem
- Converted migration `2024_12_04_073048_make_files_private.php` to use the new filesystem migration class
- Added unit tests to ensure the introduced safeguards and functions are working properly


app/Console/Commands/MakeMigration.php
Severino, line 1: I made the command overwrite the default command, 
as I would expect everybody want to turn off the activityLogging by default.
And only in special cases the user can explicitly decide to implement a 'classic migration'.

What do you think?

app/Migration/Migration.php
Severino, line 1: I added the base class so that the logging doesn't need to be taken into account for all tests.
The downside is that the child classed must use the: migrate and rollback functions to implement the migrations,
as the up and down is handled by the Base Class.

Additionally, I conciously didn't change any existing migrations (only the private files one) as I thought we want to touch production migrations. Then this would be only relevant for future migrations.

app/Console/Commands/MakeMigration.php
v1r0x: I don't really like to overwrite core artisan commands. Might be confusing for users from different laravel projects. I'd suggest using `Illuminate\Database\Events\MigrationStarted` and `Illuminate\Database\Events\MigrationEnded` events that get fired before and after a migration is called. I'll push my proposal.

app/Migration/Migration.php
v1r0x: Using my approach with `Illuminate\Database\Events\MigrationStarted` and `Illuminate\Database\Events\MigrationEnded` events we no longer have to deal with that.

v1r0x:
Migrations should now have logging disabled in all migrations. I also moved the FilesystemMigration logic from its own class to a trait. So users can now simply add `use FilesystemMigration;` to their migration, which allows to use `$this->safelyMoveDirectoryBetweenDisks()` and automatically check for `ALLOW_FILESYSTEM_MIGRATIONS`.

UnitTests are working fine, but please re-review :wink:

Severino:
This looks nice. As I understand, the logging now gets disabled in any scenario.
The question is whether there is a scenario (e.g., activity log) where we actually want activity to be logged.

Thinking about it, it should be as simple as in the 'old' migrations by just reenabbling it at the beginning of the migration:
activity()->enableLogging();

v1r0x:
Yes, logging in migrations is now always disabled. The events get the migration instance, but I couldn't find a way to get any data from that. This would be the best way to add an option to re-enable logging I think.

Edit: Maybe simply calling `activity()->enableLogging()` in the actual migration would be enough, have you already tested it?

PR #583, opened 2025-09-17
[0.12] Restrict user access using Accesspoints

v1r0x:
Accesspoints allow admins to restrict access for users to certain parts of the application. Currently only a _Default_ accesspoint is available, which redirects to the Spacialist main page. If no accesspoints are set for a user, the _Default_ is used.
Plugins can extend the list of available accesspoints by defining them in their `info.xml` file:

```xml
<info>
...
<accesspoints>
    <accesspoint>
        <path>/subpath</path>
        <id>UniqueForPlugin</id>
        <label>key_in_i18n_json_files</label>
    </accesspoint>
</accesspoints>
...
</info>
```


Severino:
Please supply your test plugin that I can use for testing.

database/migrations/2025_06_24_134012_add_user_accesspoints.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        activity()->disableLogging();
```
Severino, line 14:
```
activity()->disableLogging();
activity()->enableLogging();
```
Wrapping migrations like this is now (on target branch) handled by the DisableMigrationLogging.
We have a lot of other migrations using this; we could refactor those all together.

v1r0x:
Already uploaded :wink:

app/Utils/AccesspointUtils.php
Severino, line 1: IMO: This is more the base "model'' for the Accesspoint class. I intended the Utils folder more for general-purpose utilities like "Numbers" and "String" Utilities. 

What do you think about moving this to a designated "User" folder. Other User functionality can be moved there in the future.

resources/js/api.js
```diff
@@ -22,6 +22,10 @@ export async function getCsrfCookie() {
     }));
 }
 
+export async function refreshSession() {
+    return $httpQueue.add(() => http.get('/refresh'));
+}
+
```
Severino, line 30: This is currently not used use and document usage or remove.

resources/js/bootstrap/plugins.js
```diff
@@ -198,6 +208,34 @@ export const SpPS = {
         }
         SpPS.data.plugins[options.id] = mergedOptions;
     },
+    registerAccessPoint: (options) => {
+        if(!options.id) {
+            throw new Error('Your plugin needs an id to be installed!');
+        }
```
Severino, line 222: Redundant plugin validation.

resources/js/bootstrap/stores/system.js
```diff
@@ -128,25 +148,42 @@ export const useSystemStore = defineStore('system', {
...
 
-            const attributeStore = useAttributeStore();
-            const bibliographyStore = useBibliographyStore();
-            const entityStore = useEntityStore();
+            return true;
+        },
+        async setUser() {
             const userStore = useUserStore();
 
             const userData = await fetchUser();
```
Severino: This looks like it belongs in the user store?

resources/js/components/EntityDetail.vue
```diff
@@ -317,7 +317,7 @@
                         v-if="state.attributesFetched && !group.hidden"
                         :ref="el => setAttrRefs(el, group.id)"
                         v-dcan="'entity_data_read'"
-                        class="h-100 overflow-y-auto row"
+                        class="h-100 overflow-y-auto row mt-3"
```
Severino: This breaks the layout:
image

resources/js/components/AppView.vue
```diff
@@ -0,0 +1,595 @@
...
+            onMounted(async _ => {
+                if(!systemStore.checkAccess(currentRoute.matched[0].path)) {
+                    return;
+                }
+
+                systemStore.initialize(locale).catch(e => {
+                    if(e.response.status == 401) {
+                        systemStore.setAppState(true);
+                    } else {
+                        throwError(e);
```
Severino, line 569: Missing import

Severino (dismissed):

database/migrations/2025_06_24_134012_add_user_accesspoints.php
v1r0x: Yes, let's do this before final 0.12 release

resources/js/api.js
v1r0x: In one of the plugins I have a logout timer. If the timer is reset, it should also refresh the session.

resources/js/bootstrap/plugins.js
v1r0x: I added this because on the next lines there is a check if a plugin with that ID is already installed to prevent users from register an accesspoint before registering the plugin itself.

resources/js/bootstrap/stores/system.js
v1r0x: You are right, `setUser` should be in the user store. `checkAuthState` and `checkAccess` are IMO more system related (even though they rely on user data). But I'm ok with moving it to the user store if you think it belongs there.

resources/js/components/EntityDetail.vue
v1r0x: I replaced it with `pt-3` because not all attributes already have a padding on y-axis (e.g. Textbox)

resources/js/bootstrap/stores/system.js
Severino: Currently the program logic is this:
```js
//checkAuthState()
await **fetchUser** ()
// setUser
const userStore = useUserStore();
await **fetchUser**()
...
```

If we move it to the user store we can stick with `checkAuth` to do something like this:
```js
...,
async checkAuth(){
    try{
        const user = await this.fetchUser()
        setActiveUser(user)
        return user
    } catch {
        return null;
    }
}

File: app/Utils/AccesspointUtils.php
Comment (Severino) [reply app/Utils/AccesspointUtils.php:1]: Thinking about it it seems more like a plugin class. My suggestions would be:
a) just move it inside the plugin.php
b) we create an app/Plugin directory and add an 
- (Plugin)Accesspoints.php
- (Plugin)Scopes.php

File: app/Utils/AccesspointUtils.php [hunk 7]
```diff
@@ -0,0 +1,26 @@
...
+            "path" => "/",
+        ],
+    ];
+
+    public static function get(): array {
+        $accesspoints = self::CORE_ACCESSPOINTS;
+
+        $installedPlugins = Plugin::getInstalled();
+        foreach($installedPlugins as $plugin) {
+            $plugin->getScopes();
```
Severino: Correct me if I'm wrong. But this doesn't do anything, right? 
The returned scopes are not used.

resources/js/api.js
Severino: I kinda don't like an 'abandoned' function for potential plugins, when they can use it in a one-liner. 
Maybe it's useful at some point. I have no strong feelings about it, but at least add a comment that we don't remove it at a later point in time.

app/Utils/AccesspointUtils.php
v1r0x: Yes, this is not meant to be there.

resources/js/bootstrap/stores/system.js
v1r0x: I can't remember, why separating the same logic into to functions. If your approach is working: go for it :+1:

app/Utils/AccesspointUtils.php
v1r0x: Accesspoints are not necessarily plugin-related. Moving it to app/Plugin or Plugin.php makes no sense. Something like Utils or Helpers are my preferred "parents".

resources/js/bootstrap/plugins.js
Severino: No that's fine:

```js
    if(!options.id) {
       throw new Error('Your plugin needs an id to be installed!');
   }
```   
This is just duplicated and could be put in a separate function.
 And useful for future functions.

app/Http/Controllers/HomeController.php
```diff
@@ -103,6 +143,12 @@ public function getGlobalData() {
         ]);
     }
 
+    // Might be used in Core or Plugins to refresh a session with a short TTL
+    // e.g. for timer-based logout
+    public function updateSession() {
+        return response()->json();
+    }
+
```
Severino: Seems like a UserController function.
To update the user's session.

app/Utils/AccesspointUtils.php
Severino: You're right. I'm thinking about it.

resources/js/components/EntityDetail.vue
Severino: looks good.

app/Globals.php
```diff
@@ -50,4 +50,20 @@ public static function getGeometryTypes(): array {
...
+
+        $installedPlugins = Plugin::getInstalled();
+        foreach($installedPlugins as $plugin) {
+            $pluginScopes = $plugin->getScopes();
+            if(array_key_exists($modelClass, $pluginScopes)) {
+                foreach($pluginScopes[$modelClass] as $scope) {
+                    $scopes[] = $scope;
+                }
+            }
+        }
```
Severino: This also looks like a Plugin function. 
Do we need a Globals::getModelScopes ? 
Wouldn't the core scopes not be implemented in code? 
If this isn't necessary, I would remove this and move it to e.g. Plugin::getScopes

app/Entity.php
```diff
@@ -98,6 +98,13 @@ public static function getSearchCols(): array {
         return array_keys(self::searchCols);
     }
 
+    protected static function booted(): void {
+        $pluginScopes = Globals::getModelScopes(self::class);
+        foreach($pluginScopes as $pluginScope) {
+            static::addGlobalScope(new $pluginScope);
+        }
```
Severino: If the Globals function is moved to plugins this is fine.
Otherwise please rename to modelScopes (wouldn't be Plugin exclusive then).

app/Globals.php
Severino: 👍

app/Utils/AccesspointUtils.php
Severino: -> was moved to Service Class AccessPointsService

resources/js/api.js
Severino: Comment was added 👍

database/migrations/2025_06_24_134012_add_user_accesspoints.php
Severino: 👍Will be done before release 0.12.

Created DEV Sache §239 for this.

resources/js/bootstrap/stores/system.js
Severino: Okay, then I applied this in:
c450d20a917f48690761b39f7822785f27d5e2a3

app/Http/Controllers/HomeController.php
Severino: 👍

app/Entity.php
Severino: I moved this to a Trait:
HasPluginScopes

Severino:
I would like you to document the accesspoint functionality in the docs (I already created the specific sections for it):
image
https://github.com/DH-Center-Tuebingen/spacialist-docs/blob/main/docs/user/user-and-role-management.md
https://github.com/DH-Center-Tuebingen/spacialist-docs/blob/main/docs/developer/plugins.md

Then I have no further objections to merge this.

v1r0x:
> I would like you to document the accesspoint functionality in the docs (I already created the specific sections for it): https://github.com/DH-Center-Tuebingen/spacialist-docs/blob/main/docs/user/user-and-role-management.md https://github.com/DH-Center-Tuebingen/spacialist-docs/blob/main/docs/developer/plugins.md
> 
> Then I have no further objections to merge this.

I started working on the texts, but it sure needs some polishing.

PR #584, opened 2025-10-14
[0.11.2] Custom XSRF Token Name

Severino:
With our own structure hosting ThesauRex and Spacialist in subfolders of our domain.
Currently I experience the user, when login into one application, being logged out of any other application.

I already addressed this by storing the cookies with the respective subpath and prefixing the cookies by instance name.
But so far the problem persists. This PR tries to rename the XSRF Cookie depending on app name.

--- 

Solved the problem of XSRF-Token mismatch vs. getting logged out of the applicataion (required similar changes in Thesaurex)

1. Remember Token needs to be at root level path so that the same token can be updated on any request. (Is this even required with sanctum anymore? it feels outdated when using sanctum.)
2. XSRF-Token could be at a different path, but we need the remember token at project root, so setting the app.path would not be possible. Therefore the XSRF-Token is renamed to `XSRF-TOKEN-APP-NAME-SPACIALIST`
3. session token is renamed to `app_name_spacialist_session`

Has the benefit that we are automatically logged in Thesaurex and Spacialist.

**THE APP_KEY MUST BE THE SAME FOR BOTH APPS!**


Severino:
Updated description for the final solution.

v1r0x:
I'd like to get rid of simple imports from external libraries. Thus I replaced lodash's `trim` with our own trim function

Severino:
When we use lodash anyways, why not use the utility?

resources/js/helpers/filters.js
```diff
@@ -231,4 +231,12 @@ export function mentionify(value) {
         newValue = newValue.replace(replRegex, tpl.replace('{name}', name));
     }
     return newValue;
+}
+
+export function trim(str, charlist) {
```
Severino: why is it a filter?

v1r0x:
> When we use lodash anyways, why not use the utility?

IMO the less external functions, the better. Also a tiny step to get rid of all imports 😉

resources/js/helpers/filters.js
v1r0x: 🤷 didn't know where it fits the best. Kinda feels like a filter, but might also fit into helpers.js :D

Severino:
Then it's okay.

resources/js/helpers/filters.js
Severino: "helpers/string.js"

filters is really weird, as there are functions for Date, SiUnit (probably from dimension attr), bytes, String utilities like toFixed, highlight, bibtex utilities.

I think it would be good to separate those in the long run. Same for helpers.js.

PR #585, opened 2025-10-15
[0.11.2] Fix attribute errors on save

Severino:
I received error while writing data. 

Arrays were not treated correctly and resulted in errors. 

1. Removed business logic from controller
2. Moved Patch logic into Entity.php and AttributeValue
3. Added 'changed_attributes' to response
4. Simplified update logic in the frontend
5. Fixed error in the removedData logic


Severino:
This is rebased for the XRSF Cookie fiex. Merge that first.

PR #588, opened 2025-10-29
[0.12] Moved Plugin Scopes To Trait, Added Caching And Added TestCase 

Severino:
- [Replaced manual file path with plugin methods](https://github.com/DH-Center-Tuebingen/Spacialist/commit/1fa1c438b8688e5864ce770cb9d0b25d10eb2e65)
- [Added scope test plugin](https://github.com/DH-Center-Tuebingen/Spacialist/commit/64d736daffdb85d8ceda8c35edf284d37f9210e1)
- [Moved PluginScopes to a separate trait](https://github.com/DH-Center-Tuebingen/Spacialist/commit/c3001e5f987a3893a1d832d43382264f54a24fe3)
- [Added test case for scopes](https://github.com/DH-Center-Tuebingen/Spacialist/commit/bd1dd54173db537ff9336122e91015effa8abaab)
- Fixed 2 Failing Test Cases

### DIFF
--- a/app/Entity.php
+++ b/app/Entity.php
@@ -7,6 +7,7 @@ use App\AttributeTypes\SqlAttribute;
 use App\Exceptions\AmbiguousValueException;
 use App\Import\EntityImporter;
 use App\Traits\CommentTrait;
+use App\Traits\HasPluginScopes;
 
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
@@ -25,6 +26,7 @@ class Entity extends Model implements Searchable {
     use CommentTrait;
     use SearchableTrait;
     use LogsActivity;
+    use HasPluginScopes;
 
     /**
      * The attributes that are assignable.
@@ -69,7 +71,7 @@ class Entity extends Model implements Searchable {
         // 'entity_type_id'   => 'integer|exists:entity_types,id',
         // 'root_entity_id'   => 'integer|exists:entities,id',
     ];
-
+    
     public function getActivitylogOptions(): LogOptions {
         return LogOptions::defaults()
             ->logOnly(['id'])
@@ -98,12 +100,6 @@ class Entity extends Model implements Searchable {
         return array_keys(self::searchCols);
     }
 
-    protected static function booted(): void {
-        $pluginScopes = Plugin::getScopesFor(self::class);
-        foreach($pluginScopes as $pluginScope) {
-            static::addGlobalScope(new $pluginScope);
-        }
-    }
 
     public static function getFromPath($path, $delimiter = "\\\\"): ?int {
         if(!isset($path)) {
--- a/app/Entity.php
+++ b/app/Entity.php
@@ -71,7 +71,7 @@ class Entity extends Model implements Searchable {
         // 'entity_type_id'   => 'integer|exists:entity_types,id',
         // 'root_entity_id'   => 'integer|exists:entities,id',
     ];
-    
+
     public function getActivitylogOptions(): LogOptions {
         return LogOptions::defaults()
             ->logOnly(['id'])
--- a/app/Plugin.php
+++ b/app/Plugin.php
@@ -39,19 +39,19 @@ class Plugin extends Model
         'licence',
         'title',
     ];
-    
+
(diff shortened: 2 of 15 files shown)


app/Entity.php
```diff
@@ -98,12 +100,6 @@ public static function getSearchCols(): array {
         return array_keys(self::searchCols);
     }
 
-    protected static function booted(): void {
-        $pluginScopes = Plugin::getScopesFor(self::class);
-        foreach($pluginScopes as $pluginScope) {
-            static::addGlobalScope(new $pluginScope);
-        }
-    }
```
Severino, line 106: This was moved to `HasPluginScopes.php` then we can add the trait to every model we want in the future.

app/Plugin.php
```diff
@@ -170,51 +171,58 @@ public function getAccessPoints(): array {
...
-                    $className = Str::replaceEnd('.php', '', $src);
-                    $namespacedSrc = "App\Plugins\\$this->name\Scopes\\$className";
+    private function getScopeCacheKey(): string {
+        return 'plugin_scopes_' . $this->id;
+    }
 
-                    if(!array_key_exists($on, $scopes)) {
-                        $scopes[$on] = [];
+    public function getScopes(): array {
+        return Cache::rememberForever($this->getScopeCacheKey(), function() {
```
Severino, line 179: I added the Cache to make access faster without the need to parse the info.xml on every request.
It will be invalidated on install, update, uninstall and remove of the respective plugin.

app/Plugin.php
```diff
@@ -343,12 +351,17 @@ public function updateUpdateState($fromInfoVersion): void {
             $this->save();
         }
     }
+    
+    public function clearCache(): void {
+        Cache::forget($this->getScopeCacheKey());
```
Severino, line 355: Here we can add all other plugin Caches, e.g. PluginHooks

composer.json
```diff
@@ -44,13 +44,15 @@
...
+            "App\\Plugins\\": "app/Plugins/",
             "Database\\Factories\\": "database/factories/",
             "Database\\Seeders\\": "database/seeders/"
         }
     },
     "autoload-dev": {
         "psr-4": {
-            "Tests\\": "tests/"
+            "Tests\\": "tests/",
+            "App\\Plugins\\": "tests/assets/Plugins/"
```
Severino, line 55: In tests we use this directory. Which was not detected by the auto loader when adding the scopes to the test plugins.
With this the tests should run without problems.

database/seeders/Testing/PluginSeeder.php
```diff
@@ -35,6 +35,16 @@ public function run() {
...
+            2 => array(
+                'id' => 3,
+                'name' => 'ScopePlugin',
+                'version' => '3.2.0',
+                'uuid' => '123e4567-e89b-12d3-a456-426614174004',
+                'update_available' => null,
+                'installed_at' => null,
+                'created_at' => Carbon::createFromFormat('Y-m-d H:i:s', '2020-08-01 08:00:00', 'UTC'),
+                'updated_at' => Carbon::createFromFormat('Y-m-d H:i:s', '2020-08-01 08:00:00', 'UTC'),
+            ),
```
Severino, line 47: Added the scope plugin to have a plugin that provides a scope.

tests/TestCase.php
```diff
@@ -87,4 +87,23 @@ public function userRequest() {
...
+        
+        // Clear global scopes
+        $scopesProperty = $reflection->getProperty('globalScopes');
+        $scopesProperty->setAccessible(true);
+        $scopesProperty->setValue(null, []);
+        
+        // Flush event listeners and reboot
+        $modelClass::flushEventListeners();
+        $modelClass::boot();
+    }
```
Severino, line 108: Somehow PHPUnit behaves a little bit different than PHP in production. In Production the whole program is booted from ground up. In the unit test it remains booted during the process. So when installing a plugin, it would not be registered as a new boot, breaking the scope logic. 

Calling this with the model class reboots the model class manually

Severino:
The failing tests are from patchAttributes which should be fixed in #585

v1r0x:
:+1:

PR #589, opened 2025-10-29
[0.12] Required Attributes

v1r0x:
This PR adds the possibility to flag entity attributes as required. This require users to fill in required attributes before saving an entity.
I added `isFormValid()` to EntityDetail.vue as function, because it didn't work as computed prop. It has an optional parameter (default is true) to show a toast, if it fails because of missing required attributes.


resources/js/components/EntityDetail.vue
```diff
@@ -838,6 +847,56 @@
...
+                        &&
+                        (
+                            !hasKey(dirtyValues, attribute.id)
+                            ||
+                            isEmpty(dirtyValues[attribute.id])
+                        )
+                    ) {
+                        isValid = false;
+                        break;
+                    }
```
Severino: Move to meaningful helper file.

resources/sass/_style.scss
```diff
@@ -115,7 +115,7 @@ dl.dl-0-mb > dd:last-child {
     padding: 0;
     height: 1.8rem;
     width: 1.8rem;
-    background-color: $white;
+    /* background-color: $white; */
```
Severino: remove

Severino (changes requested):
Required is not checked in the backend. Making the required option only a fronend gimick. 
I would assume the backend has to verify that the required options are met.

resources/js/components/AttributeList.vue
```diff
@@ -109,6 +125,12 @@
                                     :title="translateConcept(element.thesaurus_url)"
                                 >
                                     {{ translateConcept(element.thesaurus_url) }}
+                                    <span
+                                        v-if="element.pivot?.metadata?.required"
+                                        class="text-danger"
+                                    >
+                                        <i class="fas fa-fw fa-xs fa-asterisk align-top" />
+                                    </span>
```
Severino: This gets pushed underneath the attribute and is therefore not visible on long labels (that get truncated).

resources/js/components/EntityDetail.vue
v1r0x: This heavily relies on live data. Move this to a helper file with a lot of parameters feels wrong.
Maybe adding local functions (hasRequiredValueRemoved, hasRequiredValueEmpty) is a better solution?

v1r0x:
> Required is not checked in the backend. Making the required option only a fronend gimick. I would assume the backend has to verify that the required options are met.

Good point :grin: :sweat_smile:

v1r0x:
@Severino can we merge this?

resources/sass/_style.scss
```diff
@@ -115,7 +115,6 @@ dl.dl-0-mb > dd:last-child {
     padding: 0;
     height: 1.8rem;
     width: 1.8rem;
-    background-color: $white;
```
Severino, line 118: This was necessary, otherwise the fab icons inside the data model editor's attribute list will show the underlying text. Either reenable or give the fab-button group a white bg as in the entity type list.

resources/js/components/EntityDetail.vue
```diff
@@ -838,6 +847,56 @@
             const hideHiddenAttributes = _ => {
                 state.hiddenAttributeState = false;
             };
+            const isFormValid = (grps, asToast = true) => {
+                let isValid = true;
+                const dirtyValues = getDirtyValues(grps);
+
+                for(let attributeId in state.requiredAttributes) {
```
Severino: This is incorrect. Here you receive the array index of the attributes not the id, should be:
```js
for(let attribute of state.requiredAttributes) 
```

resources/js/components/EntityDetail.vue
Severino, line 820: This seems unnecessarily complicated, why can't we just use this:

```js
                for(let attribute of state.requiredAttributes) {
                    // If the value is modified we need to check the dirty value, otherwise we need to check the existing value
                    const dirtyOrExistingValue = (dirtyValues[attribute.id]) ? dirtyValues[attribute.id] : state.entity.data[attribute.id]?.value;

                    if(isEmpty(dirtyOrExistingValue)) {
                        isValid = false;
                        break;
                    }
                }
```

resources/js/components/DataModelDetailView.vue
```diff
@@ -264,12 +265,36 @@
...
+                        entityAttributeId,
+                        metadata,
+                    ).then(_ => {
+                        if(!attribute.pivot.metadata) {
+                            attribute.pivot.metadata = {};
+                        }
+                        attribute.pivot.metadata.required = isRequired;
+                    });
+                }
+            };
```
Severino: Can't we just handle this via the store?
Normally the true value is inside the store, then we shouldn't have to update a value inside a prop.

Imo, it should be as easy as:
```
entityAttributeStore.setRequired(element.id, required)
```

Additionally:
It could also be executed directly on the attribute list. 
As the fab controls will forever have this single purpose. 
Those could be put into a slot of the AttributeList, but separating those from this 
logic does not make much sense.

Severino (changes requested):
It's not working for me.

I found the cause and other smaller issues.

resources/js/components/EntityDetail.vue
v1r0x: Is this really a problem? As far as I see is the only problem that that variable is `attributeId` instead of something like `requiredArrayIndex`. I only reference `attributeId` to get the actual attribute (`const attribute = state.requiredAttributes[attributeId]`). From that on I only reference `attribute.id`. This should be the same as your approach.
Nevertheless is the variable name confusing.

resources/js/components/EntityDetail.vue
v1r0x: I have to re-check. But if I remember correctly, it needs these two different checks, because of all the different states.

v1r0x: Your ternary condition is wrong.
For the first (value deleted) check we explicitly have to check for existing value in the entity data and an existing, but empty value in dirty values. Your code falls back to the existing value and thus allow deleting data
The second check seems to be identical from the logic, but I prefer the "verbose" check so it is obvious.

Severino: Yeah, you are right. But still I would prefer a more readable solution. What really bugged me was the 
if(**id** & ....)
if(**!id** & ....)

Which suggests that it's more an if/else situation. But working on it I also used a different approach.

```js
for(let attribute of state.requiredAttributes) {

    const dataEntryExists = Boolean(state.entity.data[attribute.id]?.id);
    const dirtyRequiredValue = dirtyValues[attribute.id];

    const deletingRequired = dataEntryExists && hasKey(dirtyValues, attribute.id) && isEmpty(dirtyRequiredValue);
    const requiredNotSet = !dataEntryExists && hasKey(dirtyValues, attribute.id) && !isEmpty(dirtyRequiredValue);

    if(deletingRequired || requiredNotSet)
    {
        isValid = false;
        break;
    }
}
```

resources/js/components/EntityDetail.vue
Severino: As you say. It's not really a problem during execution, but the name is still incorrect. 

It didn't work for me at all while testing and the wrong naming required additional debugging until I realized that it's not the attributeId I'm working with. Just JavaScript things :)

v1r0x: I switched to your `for … of` approach, because the index is not important and direct access to attribute is much easier.

v1r0x:
Everything done

app/AttributeValue.php
```diff
@@ -147,6 +156,11 @@ public static function handlePatch(int $entity_id, int $attribute_id, mixed $val
                 $error = __('Unknown operation');
         }
 
+        if($operation != 'remove' && !isset($value)) {
+            $error = __('Required attribute is missing.');
+            $code = 422;
+        }
+
```
Severino, line 163: Shouldn't we do this at the top and return early?
I checked the callee and there we use it inside a transaction, 
but we cannot guarantee that someone calls this outside of 
a transaction.

Severino (changes requested):

Severino:
I have fixed a small style problem that made the asterix disappear when hovered. 
[373261b](https://github.com/DH-Center-Tuebingen/Spacialist/pull/589/commits/373261bf8fffa4fdb10be18bed94ea70a1d2f335)

Severino:
Optional: I added the actual missing field names to the error message in this pull request.
Merge or Delete.

#599

app/AttributeValue.php
v1r0x: Yes, this could maybe done earlier.
I had to merge this by hand, because of diverged branches (again). If the logic of `$value` and `$attrval` didn't change, it's safe to move it to the top. Otherwise this might have to be `!isset($attrval)` instead of `$value`.

Severino: I think I was wrong. There is no saving happening before this. Should be fine. firstOrNew just creates the builder.

I think the attrval does always exist (as it is created when it doesn't exist), Checking for `isset($value)` is more a hopium approach when we e.g. get an empty object, it would also pass. 

However, currently there is no reasonable way to do it. Each Attribute would require an implementation of 'isEmpty'. That we check against when doing the `getFormattedKeyValue`.

But I think it's good enough for the moment.

Severino:
From my site this is ready to merge @v1r0x

PR #590, opened 2025-10-30
[0.12] Moved the AccesPointUtils into a Service.

Severino:
The AccessPoints Logic feels weird, as it's more part of the business logic than a universal utility class.
After some digging, I found out, that business logic is primarily managed in [Service Classes.](https://laravel.com/docs/12.x/container)

I used to put the most logic into the Models but that seems to be not such a great idea,
as the models should only be used for data storage.

So when not to put the logic in the Controller and not in the Model, Services are a nice alternative,
that let's us separate logic from the request and the controller.

I argue, we should keep that in mind and as best practice for the future.


app/Http/Controllers/HomeController.php
```diff
@@ -28,7 +28,7 @@ class HomeController extends Controller
      *
      * @return void
      */
-    public function __construct()
+    public function __construct(private readonly AccessPointsService $accessPointsService)
```
Severino, line 31: in Controllers (not in Models) those services can be injected like this. Making them accessible with $this->accessPointsService

app/Plugin.php
```diff
@@ -353,6 +354,10 @@ public function updateUpdateState($fromInfoVersion): void {
 
     public function clearCache(): void {
         Cache::forget($this->getScopeCacheKey());
+        
+        // TODO: Models are meant for the data layer only, 
+        // we should restructure the code into service classes (e.g. PluginMigrationService) 
+        app(AccessPointsService::class)->clearCache();
```
Severino, line 360: This is a bit ugly, but see comment.

app/Plugin.php
```diff
@@ -484,7 +489,7 @@ private function publishScript(): void {
 
             self::getDirectory()->store(
                 $this->publicName(false),
-                $scriptPath
+                $filehandle
```
Severino, line 492: I introduced a severe bug, that the file did not write the js file but the file path to the file. 
This fixes it again.

app/Services/AccessPointsService.php
Severino, line 1: This is basically the AccespointUtils, just in a Service Container.

app/Services/AccessPointsService.php
```diff
@@ -0,0 +1,39 @@
...
+    ];
+    
+    public function clearCache(): void {
+        Cache::forget(self::CACHE_KEY);
+    }
+
+    public function get(): array {
+        return Cache::rememberForever(self::CACHE_KEY, function() {
+            $accesspoints = self::CORE_ACCESSPOINTS;
+            $accesspoints = array_merge($accesspoints, self::loadAccessPointsFromPlugins());
```
v1r0x: Shouldn't this be `$this->loadAccessPointsFromPlugins()` instead of `self::loadAccessPointsFromPlugins()`?

Severino: Yeah you are right. I'll fix it.

Weird part is, I tested it and it was working. 
In PHP it seems this is working, too 😶
```php 
<?php
// example code
class SelfTest {

    public function do(){
        self::stuff();
    }

    public function stuff(){
        echo "STUFF HAPPENDED";
    }
}


new SelfTest()->do();
// Prints "STUFF HAPPENDED";
```

v1r0x: Yes, this is a relict from PHP < 5 that did not have a static keyword. But: this should raise a deprecation warning in PHP8 and throw an error in PHP9.
But my IDE did not show any warnings :shrug:

v1r0x (approved):
:+1:

PR #594, opened 2025-12-09
[0.13] Migrate To PEST As Testing Framework And Add Browser Testing

Severino:
- Used auto-migration to move unit tests to the PEST framework and fixed errors not handled by auto-migration
- Added PEST v4 browser testing functionality
  - PHP version requirements changed to: 8.3 and 8.4,because PEST browser testing requires PHP 8.3 (8.2 should still work but is no longer covered by tests)
  - Added basic test case for site access
  - Added test case for login page

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -1,6 +1,14 @@
 name: PhpUnit
 
-on: ["push", "pull_request"]
+on:
+  push:
+    branches:
+      - main
+      - 'release-**'
+  pull_request:
+    branches:
+      - main
+      - 'release-**'
 
 jobs:
   laravel-tests:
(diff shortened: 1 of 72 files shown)


Severino:
@v1r0x I tested the migration and adding of PEST for backend and frontend testing. 

In the end, PEST just makes the written tests shorter, more concise and provides better readability. But still uses PHPUnit. 
The browser testing capabilities look sufficient and promising. Then we would have all testing under one umbrella.

We can talk tomorrow about it, if we want to proceed this way.

v1r0x:
Having one easy to use and maintainable testing framework sounds good. But I think we should focus on releasing 0.12 and merging all open PRs and cleanup outdated branches/ideas.
Then, a **fast** switch to PEST can be a first step for 0.13 (or whatever comes after 0.12). This makes it easier to finish 0.12 without having to deal with conflicts due to updated/new tests.

PR #599, opened 2026-03-19
Prints list of missing required attributes to the user

Severino:
For debugging I added the missing required attributes to the error message for convenience.
If you think it's good, merge otherwise delete this branch.

required-message

### DIFF
--- a/resources/js/components/EntityDetail.vue
+++ b/resources/js/components/EntityDetail.vue
@@ -796,8 +796,8 @@
                 state.hiddenAttributeState = false;
             };
             const isFormValid = (grps, asToast = true) => {
-                let isValid = true;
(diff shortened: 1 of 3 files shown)


v1r0x:
I'm not sure if it really helps or "overwhelms" the user if there are more than 4-5 missing attributes or even less with longer names.
In that case the toast has to be much wider or higher and takes a lot of space or we have to cut the list after the first two missing attributes and add a `and x more`. Both solutions are not ideal. I'd wait for user feedback and rework it in the future.

Severino:
I mean normally it's just when you miss something and then it's more like 1-2.
Also when the required attribute is on another tab its suboptimal, but at least has some information.

 But nvm was just useful for debugging.

PR #601, opened 2026-03-19
[0.12] Allow multi move using the multi select capability

Severino:
updated version #506

Allows to move multiple entities with the 'Right-click -> Move' functionality when in the existing 'Multi-edit-selection' mode. 
I consider this very useful and would not want to scrap it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Plugin-System now supports PluginScopes
 - .env variable `ALLOW_FILESYSTEM_MIGRATIONS` to explcitly enable filesystem migrations
 - API endpoint for getting all entity details data in one request: GET::v1/entity/{id}/entity_detail
+- Multi move in entity tree: Select multi edit mode, select multiple entities in the tree and then right click 'move' to move all selected entities at once
 ### Fixed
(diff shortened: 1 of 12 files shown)


Severino:
Merge #600 first

v1r0x:
I fixed a couple of things as the behaviour was pretty weird.

1. use `patch` instead of `post` to match the single entity move api endpoint (also you added it to the patch section in `api.php`
2. Rename api endpoint to prevent camel-case in endpoints
3. Do not show entity name in move modal if multiple entities are selected
4. Disable selection non-matching entities in selection mode to prevent differences in what is selected in the UI and what is actually selected -> leads to missing alert in move modal
5. Disable _Move_ dropdown item if either a non-selected entity is selected or entities of more than one entity type are selected. It makes no sense to allow users to click a dropdown entry and then tell them that they can not execute this operation

v1r0x (approved):
Ready to merge @Severino but please review my changes

Severino:
Thank you, that's very good. Esp. the entity type check.

I removed the comment and used meaningful variable names in the 'disabledMenuEntries'. And moved the transparency, when an entity cannot be selected, to the whole node which makes the disabled state more visible.

v1r0x (approved):
👍

PR #602, opened 2026-03-20
[0.12] feature auto columns preferences

Severino:
Currently the user must sum columns manually in the preferences and calculate the number to 12 always in their head, requiring a big explanation text, why he has to do so. 

I implemented this feature in an old branch and think it's worth implementing (until columns can be resized freely).
auto-calculate-columns

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - Now entity metadata is only loaded when accessing the metadata tab
 - Migrations now have logging automatically disabled
 - Migrations that require filesystem changes can now be handled using the `App\Traits\FilesystemMigration` trait (call `$this->safelyMoveDirectoryBetweenDisks(...)` with `ALLOW_FILESYTEM_MIGRATIONS` set to `true`)
+- Column values in preferences are automatically recalculated to values that sum up to 12
 
 ## 0.11.1
 ### Added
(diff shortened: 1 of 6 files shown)


v1r0x:
Without any text it is even harder to understand what this preference does. "Magically" updating values to a different values is also an odd behaviour.

I'd prefer a whole new approach.

Severino:
I think it's a good intermediate solution that already exists until we can allow a free resizing of the columns.

We could also keep the arbitrary values and do the calculation in the background when rendering the page, but it could lead to weird and not understandable behavior for the user. E.g. one column is set to 0 but is not 0 in the column. 

Otherwise we can discuss this in the next spacialist jour fix.

v1r0x:
For me it is not a intermediate solution. Just a slightly different working UI, but with the same output. Either leave it as it is or we implement a different and better solution.

My problems with this PR are:
- We already have a working solution that everybody knows
- Your proposed solution works the same way under the hood
- I don't know any user that requested a different solution - that just works the same way
- You still have to count or be aware of fractions. Any user should be able to count to 12.
- Without any explanation it makes no difference if the sum has to be 12 or users can set it to 12000. They still have to calculate a columns width.
- Magically updating user input is (IMO) always a bad (the worst to be honest) solution. If I enter 50, 20 and 30 and end up with 6, 2, 4 I think something went wrong and wonder why it got reset to these values.

Severino:
Then let's discuss it in the group.

v1r0x:
The result of the discussion was to keep it as is and replace this with a proper resize handlebar

PR #603, opened 2026-03-20
[0.12] Move user create logic to model

Severino:
Small improvement moving data logic from the UserController to the User model.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -223,7 +223,6 @@ class UserController extends Controller {
             $request->get('accesspoints'),
         );
 
-        // TODO is this necessary? The user should already be up to date after creation [SO]
         $user = User::find($user->id);
         
         return response()->json($user);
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -215,21 +215,17 @@ public function addUser(Request $request) {
...
-        $user->save();
+        $user = User::create(
+            $request->get('name'),
+            $request->get('nickname'),
+            $request->get('email'),
+            $request->get('password'),
+            $request->get('accesspoints'),
+        );
+
+        // TODO is this necessary? The user should already be up to date after creation [SO]
```
v1r0x: Relations are not added/updated after `save()`. That's why I added this _workaround_

v1r0x: Don't know why there isn't a built-in function to do this (or why I fixed it that way instead of using a proper function). Either remove the `// TODO` and we keep it as is or replace it with a proper solution.

v1r0x (changes requested):

PR #604, opened 2026-04-16
Release 0.11.2

v1r0x:


v1r0x:
The custom XSRF code breaks user login test with wrong credentials. Please look into it @Severino

Severino:
Tests are completing again, only the Percentage is failing, as it was designed as Integer input and was changed to allow float in this PR. It should be considered if this input represents an arbitrary double value or if there is a certain precision we want to enforce. The frontend component does support floats via input or 2 places after comma with shift while dragging.

Values more detailed than .00 are not displayed very well in the component

image

Severino:
Rest seems okay.

v1r0x:
You are right, I missed to reflect these changes in the tests as well. I also added a check to the import logic to prevent importing values with a higher precision as 2

v1r0x:
I also added a couple of login related tests. Ready to merge?

PR #607, opened 2026-04-16
[0.11.2] Fix attribute value duplicates

v1r0x:


v1r0x:
The core of this branch/PR (fixing attribute value duplicates) is an important improvement. But this branch is a mess with backported 0.12 stuff. Please redo as a clean 0.12 branch with only what should be done for this fix @Severino

PR #615, opened 2026-05-29
[0.12] Plugin Restructure

Severino:
Structures all plugin features into a single PluginManager. All plugin features like (migrations, scripts, dependencies, permissions, etc) are registered while managing themselves and only reacting to the PluginManager's lifecycle calls.


Severino:
Somehow the script file is not correctly published on upload

